Issue #259, opened 2014-09-30
Checkstyle Package XML should be in core library JAR


dhumeniuk:
According to http://checkstyle.sourceforge.net/config.html#Packages, the package XML file should be in the root of the JAR providing the additional checks (sevntu-checks project) and is currently not. This would allow just the classname to be specified in the checks file like is done by the Eclipse plug-in.

See discussion here too: https://groups.google.com/forum/#!topic/sevntu-checkstyle/Krhy6UYeOAI

rnveach:
Closing as file was added

PR #260, opened 2014-09-30
Issue #259: Include package XML file in the core library.

dhumeniuk:
First PR for this repo, not sure what needs to be done, but by adding this file I was able to run checkstyle from Ant using the core library and no package information in the configuration files.


rnveach:
Closing this in favor of #516

PR #292, opened 2015-01-04
Collection with imports shall be cleared after visiting tree.

damianszczepanik:
I copied piece of code from this class and found that if list with imports is not cleared after visiting tree than invalid (additional) errors may be reported. It happened to me when I run checkstyle for project that has ~30k lines.


romani:
It will be better to do it in beginTree method
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/Check.java#L144

as if any exception during validation happen , collections will not be cleared.

romani:
@damianszczepanik, please review your PR, it has 3 commits, and not all changes are yours, please do rebase from our master.

romani:
please also make sure that commit message always reference to github issue/pr in message, see other commits for example.

damianszczepanik:
I will send this PR again.

PR #293, opened 2015-01-05
Verifies that AutoCloseable resource is declared with try-with-resources (JDK 1.7+) statement to make sure that is closed correctly

damianszczepanik:
- Added Check class for validation
- Prepared Input classes including those that work from JDK7
- Added violation to messages.properties
- added resources-noncompilance to eclipse project to simplify development

### DIFF
--- a/sevntu-checks/.classpath
+++ b/sevntu-checks/.classpath
@@ -6,7 +6,7 @@
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry excluding="**" kind="src" path="src/test/resources-noncompilable">
+	<classpathentry kind="src" path="src/test/resources-noncompilable">
 		<attributes>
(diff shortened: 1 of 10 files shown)


damianszczepanik:
OK, so I improved coverage by adding new tests and the only not coveraged code is related to exposed method reportUnsupportedToken() that reports problem with my code. Since I used http://commons.apache.org/proper/commons-io/ and own project to test the code I should consider removing this throw IllegalArgumentException and use return false.

However I'm not sure if this is right moment for this and what is the good practice. Let me know.

romani:
I do not believe such kind of validation is possible to do without knowing of types. But checkstyle is not type aware tool, all that we have are strings.

as your Check does not have any option, please put in your Input files comments like "//warn" to let easily see where your expect violations.

Check javadoc should full of details and examples as almost all other, but we can do that after I clearly see that task is doable by Chekstyle without bug amount of false positives.

PR #295, opened 2015-01-06
Improved CustomDeclarationOrderCheck

damianszczepanik:
- Added tests to increase coverage
- Fixed rule in test method
- Fixed formatting (indents)

Pull #295


damianszczepanik:
OK, let me summarize reason of changes:
1. I found that rule "MainMethod()" is not parsed and it should be "MainMethod(.*)" as it is mentioned in default order https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/CustomDeclarationOrderCheck.java#L200
Since 'main' method has only one valid signature (or rather two: with [] and ...) so using rule "MainMethod()" should ends with error. Currently both are accepted but only one is parsed.
2. Added "this" to field because one branch was not covered (expect to have DOT token)
3. Added new test to cover branch that triggers wrong "MainMethod()" order.

Finally it improves coverage by +1.18% :-)

romani:
thanks a lot !

PR #297, opened 2015-01-14
Fixed false positive in check ForbidReturnInFinallyBlockCheck. Fixes #269

### LINKED ISSUE #269: False positive: return in finally block
Warning: "Forbid return statement in finally block: Finally block should not contain return statements."

https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ForbidReturnInFinalBlockCheck.java

```
	public void test() {
		try {
                         /// some code
		} finally { 
			// remove test data
			csf.doIn(new Callback<Void>() {
				@Override
				public Void execute(ISession session) {
					session.delete(this);
					return null;  // WARNING !!!!!!
				}
			}, "");
		}
	}
```
it is also required to rename Check, as block is "finally" not a "final"

Andrew0701:


romani:
merged as FF, thanks a lot

PR #302, opened 2015-02-01
new Check UselessSuperCtorCallCheck. Fixes #186

### LINKED ISSUE #186: new Check: EmptyPublicCtorInFinalClassCheck
following code is happen : 
```
class MyClass {
 MyClass() {
   super();
 }
}
```

nobody extend this class, this class does not extend anything - its c-tor is completely useless, especially "super()";

Checkstyle idea - EmptyPublicCtorInFinalClass, 

base on investigation why empty ctor could be necessary - http://stackoverflow.com/questions/3507281/java-inheritance-problem-have-to-create-empty-constructor-in-father-class,

we can not completely cover code case that described above, as Checkstyle does not have information about other classes (nobody can predict who can extend from MyClass). 

What we can be sure is final class can not be extended - so we could catch that cases.
Nuances:
- we can check class only if it final and do not "extend" anything.
- private c-tor is OK in utility classes. 
- ignoreSuperCtorCall - Additionally check for "super();" in class's non default  c-tor and this class does not extend anything.
- even class is not final, but do not extend anything , call of "super();" is useless and should be caught by this Check.

pirat9600q:
This commit adds new Check UselessSuperCtorCallCheck.


romani:
merged as FF

PR #304, opened 2015-02-04
Improved coverage of AvoidDefaultSerializableInInnerClasses

damianszczepanik:
Improved coverage of AvoidDefaultSerializableInInnerClasses by importing full qualified class name Serializable (use java.io.Serializable)

Pull #304


rdiachenko:
This PR doesn't make sense for me because the change in the test input doesn't relate to the check's goal and it's logic (see checks' documentation and test cases) as a result it can not improve coverage for this check.

damianszczepanik:
Just o be clear - it improves coverage of the code, not test. It increases by 0.03% as mentioned above by @coveralls

rdiachenko:
Could you please point me out the piece of code for which you improved it's coverage?

damianszczepanik:
If I remember correctly this change makes below condition true and false while previously it was always false for given test data
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidDefaultSerializableInInnerClasses.java#L306

rdiachenko:
@romani I'm ok to merge

romani:
merged as FF

PR #305, opened 2015-02-04
Improved test coverity of class ChildBlockLengthCheckCheck by adding tes...

damianszczepanik:
...t for tested class

Pull #305

### DIFF
--- a/sevntu-checks/src/test/java/com/github/sevntu/checkstyle/checks/design/ChildBlockLengthTest.java
+++ b/sevntu-checks/src/test/java/com/github/sevntu/checkstyle/checks/design/ChildBlockLengthTest.java
@@ -19,7 +19,7 @@
 package com.github.sevntu.checkstyle.checks.design;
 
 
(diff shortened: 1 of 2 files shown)


rdiachenko:
Your new test case just repeats ```testManyBadChildBlocks()``` and ```testManyBadChildBlocks2()```. This check doesn't bother about the content of the blocks (whether it is inner class inside ```if``` of just other long logic).

@damianszczepanik If you don't agree please point me out the piece of code where the coverage was improved due to your changes.

damianszczepanik:
I'm not saying that given example is the minimal of the code to improve the coverity, this class could be probably simpler or shorter or even merged with existing test.

I would have to run cobertura:cobertura before and after the change to confirm but I think it was about this condition https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/ChildBlockLengthCheck.java#L206

rdiachenko:
@romani I'm ok to merge

PR #313, opened 2015-02-10
Run cobentura with mvn check or mavn install and fail if code coverage c...

damianszczepanik:
...riteria are not met

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ jdk:
   - openjdk7
   - oraclejdk7
   
-before_script: cd sevntu-checks
+before_install: cd sevntu-checks
+install: mvn clean verify
 
 after_success:
-  - mvn jacoco:report coveralls:jacoco
+  - mvn jacoco:report coveralls:jacoco 
--- a/sevntu-checks/pom.xml
+++ b/sevntu-checks/pom.xml
@@ -74,6 +74,73 @@
         </execution>
        </executions>
     </plugin>
+    <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>cobertura-maven-plugin</artifactId>
+        <version>2.6</version>
+        <configuration>
+            <check>
+                <haltOnFailure>true</haltOnFailure>
+
+                <!-- only integer values are accepted -->
+                <branchRate>100</branchRate>
+                <lineRate>100</lineRate>
+                <totalBranchRate>88</totalBranchRate>
+                <totalLineRate>97</totalLineRate>
+                <regexes>
+                    <regex><pattern>.*.checks.annotation.ForbidAnnotationCheck</pattern><branchRate>70</branchRate><lineRate>96</lineRate></regex>
+                    <regex><pattern>.*.checks.coding.AvoidConstantAsFirstOperandInConditionCheck</pattern><branchRate>80</branchRate><lineRate>100</lineRate></regex>
+                    <regex><pattern>.*.checks.coding.AvoidDefaultSerializableInInnerClasses.*</pattern><branchRate>93</branchRate><lineRate>92</lineRate></regex>
+                    <regex><pattern>.*.checks.coding.AvoidHidingCauseExceptionCheck</pattern><branchRate>87</branchRate><lineRate>100</lineRate></regex>
+                    <regex><pattern>.*.checks.coding.AvoidModifiersForTypesCheck</pattern><branchRate>96</branchRate><lineRate>98</lineRate></regex>
+                    <regex><pattern>.*.checks.coding.AvoidNotShortCircuitOperatorsForBooleanCheck</pattern><branchRate>86</branchRate><lineRate>100</lineRate></regex>
+                    <regex><pattern>.*.checks.coding.ConfusingConditionCheck</pattern><branchRate>84</branchRate><lineRate>100</lineRate></regex>
+                    <regex><pattern>.*.checks.coding.CustomDeclarationOrderCheck.*</pattern><branchRate>81</branchRate><lineRate>83</lineRate></regex>
+                    <regex><pattern>.*.checks.coding.DiamondOperatorForVariableDefinitionCheck</pattern><branchRate>90</branchRate><lineRate>100</lineRate></regex>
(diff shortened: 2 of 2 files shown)


romani:
please resolve(revert) formatting changes, I need to see funcational changes only.

damianszczepanik:
Sorry for that but I there were tabs and spaces in pom.xml file so I reformatted it. The only section that was added is the plugin at https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/313/files#diff-af4fedf1ce8aed906fb6fad9069d5193R76 so not much. Are you OK with this?

romani:
no, please redo commit, changes should not contain formatting.
you can send fomatting changes as first PR and then clean functional changes as second, choose what ever is earsier for you.

romani:
Why you have such strange values for coverage ?
mvn clean cobertura:cobertura
sevntu.checkstyle/sevntu-checks/target/site/cobertura/index.html

from Report:
Line Coverage  is about  97%
Branch Coverage is 70% minimal 

We have to enforce min coverage  by special options to make sure that build control coverage completely and does not allow any new Check to join non 100% coverage. Please read 
http://mojo.codehaus.org/cobertura-maven-plugin/usage.html and configure Cobertura to have demand 100% with some exceptions for existing Checks. 

Please open issue on sevntu.chekcstyle to update all Uts to jump to 100% in coverage and remove special rules from pom.xml

damianszczepanik:
<branchRate>50</branchRate> - the lowest brnch rate we have in other words there is file that has 50 branch coverage.
<lineRate>83</lineRate> as above but for lines
<totalBranchRate>88</totalBranchRate> - total branch rate, the value which you mentioned above
<totalLineRate>97</totalLineRate> - similar to yours

Not sure why total branch is different but what I have done was to set all values to 100 and decreased them until build passed. You may find details in previous build that Travis executed for this patch. Looks like we have different snapshot versions (I may not be rebased to head).

Cobentura does not mind if the file is new or not. I guess there is no way to tell cobentura keep 97% level for existing files and 100 new because it is not connected with git to extract information which file is new, which updated. My idea was to set values as above and review/update from time to to time coverage until the project reaches 100%.

If you want to check coverage details you can see details eg https://coveralls.io/builds/1924140 but I don't know how to trigger it in maven. Maybe you can configure it when connecting github with coveralls? Maybe it keeps history and can compare reviewed files?

romani:
> 50 - the lowest brnch rate we have ........

sevntu-checks/target/site/cobertura/index.html, open "com.github.sevntu.checkstyle.checks.naming" there is 50%
Such low testing rate is possibly my problem that I accepted new Check and did not recheck Cobertura.  That will happen again and again till we enforce it automatically

> Cobentura does not mind if the file is new or not.

sure, But in configuration (pom.xml ) we could provide certain values for each Check. So new Checks , will definitely be missed in that config, will require 100% by default. All existing Check we will slowly try to improve test coverage till 100% and remove from pom.xml any reference to custom coverage rules.
I am not 100% sure, but example http://mojo.codehaus.org/cobertura-maven-plugin/usage.html#Configuring_Check looks very promising. 

> My idea was to set values as above and review/update from time to to time coverage until the project reaches 100%

With such approach, we will never succeeded. In any code clean-up process team have to pin current status (each file have its own minimal values of coverage) and newly created code code/files should follow maximal quality threshold.

>  https://coveralls.io

that site is using jacoco library (it is alternative to cobertura), here is run command https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/.travis.yml#L10 , so some values will be different for sure. If we impose coverage from build time, we will not require coveralls anymore, travis will do all that we need.

> Maybe you can configure it when connecting github with coveralls? 

Nothing special, all details of build is in travis.yml file , I might give to coveralls my security token to receive jacoco statistics only form us (but I do not remember).

damianszczepanik:
I like the way you think. Not easy to maintain but drives us much faster to the point called 100%.

Some comments to above patch:
1. You may noticed some postfix .* for rules and this is for file that have internal classes - Cobertura does not work with such cases properly so I group them together
2. Last line contains Utils that is not covered at all but you noticed than soon this utility won't be needed any more
3. Since we expect 100% and have some exceptions I'm not sure if having and maintaining totalBranchRate and totalLineRate is really useful

romani:
1) please sqash commits in to one figure out why github show two commits in that PR.

2)
> 2. Last line contains Utils that is not covered at all but you noticed than soon this utility won't be needed any more

each line that is regexes group will be reflected in our issue tracker with deman to remove it and make coverage 100%, so Utils is not exception and shuold be covered eventually

3) 
> 3. Since we expect 100% and have some exceptions I'm not sure if having and maintaining totalBranchRate and totalLineRate is really useful

It is usefull, please see my point "2" it partly answer your question. Lets make a task to mke it 100% , for now I donot see a reason why we should not demand that.
As It is not possible to put 100% ,lets put a comment  on that lines as a reminder to make it 100% after last item is removed from "regexp" group.

4) If that approach works fine - we will move it to our main project to enforce quality there too.

damianszczepanik:
I rebased branch and it fixed problem with two patches. Is that OK/ready to merge?

romani:
Yes, now i see one commit, please add comment on totals that i requested in point 3, after that i will do merge.

damianszczepanik:
I'm not sure what comments are you expecting and where. Something like ToDo next to totalBranchRate and totalLineRate?

I have doubt if those two values make sense. The second one is exactly the same which Coveralls pins as a comment after each build so you don't need it. Also if you don't maintain those values frequently they will be soon or later outdated. Do you really want to remind everyone to update this value each time someone increase them? That would be annoying...

With current configuration (each new file must be covered not less than 100% and existing one can't be lover than declared exceptions) given metrics don't make any additional validation. In other words you can drop them both and restrictions stay the same.

romani:
ok , I did it myself.
thank you very much for your time and desire to help a project.

> Something like ToDo next to totalBranchRate and totalLineRate?

yes. I did it myself.

> Do you really want to remind everyone to update this value each time someone increase them?

No, but I need reminder to me to make them 100, after all classes are covered by 100.

PR #315, opened 2015-02-13
Improved UnnecessaryParenthesesExtendedCheck test coverage:

damianszczepanik:
- added double parentheses for assertion test
- make sure that parentheses are checked when assertion is enabled and disabled
- removed assertion for input parameter that is already checked before function is invoked (double check)


romani:
> "100:24: " + getCheckMessage(MSG_KEY_EXPR),

indentation problem, please fix.

damianszczepanik:
@romani 
Can you take a look at this?

romani:
good, but as it "test coverage" I do not see changes in pom.xml for increasing minimum coverage level :).

all improvement  should follow with improving of cobertura-maven-plugin to 100%, yes it is boring :) by it the only way gain maximal quality.

romani:
thanks a lot !

PR #316, opened 2015-02-13
Removed duplicate check for the empty list: first condition isEmpty() but then for-each which also skips iteration if list is empty

damianszczepanik:
This diff is poor but what I've changed was to remove if and changed indention

Pull #316


damianszczepanik:
Hmmm, by removing one condition that was always true, report says that I increased coverage. And the problem is that it says that number of covered lines has decreases but it does not count that the total number of lines is also decreased. So number has decreased but relation covered/all has increased.

And in that case it makes me worry about #313 - you may get fail even you improved relative coverage.

romani:
I use coverall only to keep a reminded that smb did not tested a code, on coverage decrease I always recheck UTs and update, and some time it is ok to to have that decrease , I am not a coverage fanatic but it helps to not forget about Utests.

romani:
merged as FF.

PR #320, opened 2015-02-22
improved InterfaceTypeParameterNameCheck code coverage

damianszczepanik:
Pull #320

- reduced condition that will never be true (modifiers are not direct parent for type definitions)
- added example which introduces interface as inner definition for outer class
- added example such as <T extends List>


damianszczepanik:
@romani 
I made message more compact and added some comments above. Is that correct now?

romani:
merged, but please try to make message as single line, single sentence should always be enough to describe an update. In other case it would be better to put extra details to issue/pull-request.

damianszczepanik:
I've put into the commit message only single line (header of this pull) but if this is still wrong then github might merge comments which I provided when pushing this PR to upstream. I will check and improve results!

PR #324, opened 2015-02-28
Increase AvoidNotShortCircuitOperatorsForBooleanCheck code coverage

damianszczepanik:
Pull # 324

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
@@ -115,27 +115,18 @@ public class AvoidNotShortCircuitOperatorsForBooleanCheck extends Check
     {
 
         DetailAST currentNode = detailAST;
-        while (currentNode != null
-                && currentNode.getType() != TokenTypes.EXPR
-                && currentNode.getType() != TokenTypes.METHOD_DEF
-                && currentNode.getType() != TokenTypes.CTOR_DEF
-                && currentNode.getType() != TokenTypes.CLASS_DEF)
+        // look for EXPR which is always around BOR/BAND... operators
+        while (currentNode.getType() != TokenTypes.EXPR)
         {
             currentNode = currentNode.getParent();
         }
 
-        final int type = currentNode.getType();
-
-        if (type == TokenTypes.EXPR) {
-
-            if (isBooleanExpression(currentNode)) {
-                log(detailAST, MSG_KEY, detailAST.getText());
-            }
-
-            supportedOperands.clear();
-            hasTrueOrFalseLiteral = false;
+        if (isBooleanExpression(currentNode)) {
+            log(detailAST, MSG_KEY, detailAST.getText());
         }
 
+        supportedOperands.clear();
+        hasTrueOrFalseLiteral = false;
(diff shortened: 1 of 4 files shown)


romani:
please explain each changes in logic of Check (especially testing for nulls).
Any changes in logic will require manual retesting over big projects. I am not afraid of changes in code and testing but please prove that checks for NULLs are useless.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
```diff
@@ -115,27 +115,18 @@ public final void visitToken(final DetailAST detailAST)
     {
 
         DetailAST currentNode = detailAST;
-        while (currentNode != null
```
damianszczepanik: If you pass any of operators then soon or later you get `EXPR` as parent and that's why validating against other tokens doesn't make sense. Also null value should not be possible because I imagine that `null` is not passed to `visitiToken()` method. Am I right?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
```diff
@@ -115,27 +115,18 @@ public final void visitToken(final DetailAST detailAST)
...
-                && currentNode.getType() != TokenTypes.CLASS_DEF)
+        // look for EXPR which is always around BOR/BAND... operators
+        while (currentNode.getType() != TokenTypes.EXPR)
         {
             currentNode = currentNode.getParent();
         }
 
-        final int type = currentNode.getType();
-
-        if (type == TokenTypes.EXPR) {
```
damianszczepanik: Already checked above

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
```diff
@@ -160,14 +151,12 @@ public final static boolean isBooleanType(final DetailAST node)
      */
     public final boolean isBooleanExpression(final DetailAST node)
     {
-
         DetailAST curNode = node;
 
         final List<String> childNames = getSupportedOperandsNames(curNode);
         final List<String> booleanVariablesNames = new LinkedList<String>();
 
-        while (curNode != null
```
damianszczepanik: Can you meet `null` in hierarchy between `EXPR` and any of those values? I Believe not and that's why I dropped it.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
```diff
@@ -218,7 +207,6 @@ public final boolean isBooleanExpression(final DetailAST node)
             }
 
             if (currentNode.getType() == TokenTypes.IDENT
-                    && currentNode.getParent() != null
```
damianszczepanik: if `currentNode` is `IDENT` it always has not-null value as parent.

romani:
merged as FF
make sense , thanks a lot !

PR #325, opened 2015-03-01
Moved static duplicated methods into Utils class

damianszczepanik:


damianszczepanik:
I wasn't sure if this should be added to Check or Utils but I found that generally Check does not support logic and this method is more utility so I decided to put it into Utils.

romani:
merged as FF, thanks a lot.
we resolved Utils location problem only few weeks ago in main project, before that time I tried to keep Checks as independent as possible to ease their management.

Issue #310, opened 2015-02-06
Run checkstyle checks on sources of sevntu.checkstyle


rdiachenko:
When a new check is created a developer should run his check on the existing checks sources, analyze and fix violations.

The purpose of this issue it to automate this process by the following steps:
- Update maven build configuration for `sevntu-checks` module to run all checkstyle-checks.xml from Checkstyle repositoy on sources, as that is requirement for all checks that are going to appear in main repo.
- The build should failed until all the violations will be fixed

Think about how to enforce the developer to add his new check to the list of already running checks in order so the probable violations from this new check will be fixed by developer before pull request.

damianszczepanik:
I've done it already for code coverage in https://github.com/damianszczepanik/sevntu.checkstyle/commit/a2f6734b0e3140021e56912e29fba8a264a98539 so I expect validating against checkstyle rules should work similar.

Have you prepared list of rules that shall be applied? I expect some of them are similar to Sonar but Sonar uses different approach and may check more than checkstyle (PMD, FindBugs).

We'd better start to clean #169 and then this one would be easier because we will not have exception for each class :)

romani:
general idea - https://github.com/checkstyle/checkstyle/wiki/Checkstyle-GSoC-2015-Project-Ideas#project-name-practice-what-you-preach , we failed to pass to GSoC 2015 so that idea will not be locked to one contributor, and we could start doing it.

right now we use - https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle_checks.xml config, but eventually all Checks should be added to it. We can do the same for sevntu.checkstyle.

romani:
issue is already done https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/pom.xml#L123

PR #326, opened 2015-03-01
Sonar violation: Tabulation characters should not be used

damianszczepanik:


damianszczepanik:
I found classes that use only spaces, only tabs and both. I updated only those that have mix tab and spaces.

Sorry for double push - I forgot to add id of the PR into commit message.

romani:
thanks a lot , I appreciate that !!!!!
double push is ok and does not bother me, final result is in priority.

We enforce checkstyle rules from main repo agaist this experimental project, #310 , as we have an idea to move all that Checks to main repo after testing period.

damianszczepanik:
So do you want me to apply `.replaceAll("\y", "    ")` for all repo?

#310 sounds interesting, if you can provide some tips (I added comments) I will take it. Did you already update main checkstyle repo to validate against checkstyle? I couldn't find it in .travis configuration file.

romani:
> Did you already update main checkstyle repo to validate against checkstyle?

yes, not all Checks are activated, but some of them.

>  I couldn't find it in .travis configuration file.

https://github.com/checkstyle/checkstyle/blob/master/.travis.yml#L8 in verify phase
https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L445

Issue #249, opened 2014-09-08
forbid nested Switches


romani:
Nested Switch is "good" pattern for development.

from https://www.thc.org/root/phun/unmaintain.html
```
The Nested Switch
(a switch within a switch) is the most difficult type of nesting for the human mind to unravel. 
```

Check could be done base on:
http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/AbstractNestedDepthCheck.html

romani:
assigned to @damianszczepanik

romani:
fixed.

Issue #290, opened 2014-12-30
Use cs-doclet to generate metadata files and online documentation for sevntu-checkstyle


aunkrig:
Hello SevNTU team,

I wrote a small tool called "cs-doclet" which generates
- messages.properties
- checkstyle-metadata.xml
- checkstyle-metadata.properties

from doc comments in the source code of your checks and filters, so you don't have to keep all those files in sync manually. Also, it creates human-readable web pages for each check/filter in MediaWiki markup format.

The detailed description of the tool is [here](http://unkrig.de/w/Cs-doclet).

Examples for generated files/web pages are here:
- [messages.properties](http://svn.code.sf.net/p/loggifier/code/trunk/de.unkrig.cs-contrib.core/src/de/unkrig/cscontrib/checks/messages.properties)
- [checkstyle-metadata.xml](http://svn.code.sf.net/p/loggifier/code/trunk/de.unkrig.cs-contrib.core/src/de/unkrig/cscontrib/checks/checkstyle-metadata.xml)
- [checkstyle-metadata.properties](http://svn.code.sf.net/p/loggifier/code/trunk/de.unkrig.cs-contrib.core/src/de/unkrig/cscontrib/checks/checkstyle-metadata.properties)
- [MediaWiki online documentation](http://unkrig.de/w/Cs-contrib.unkrig.de_/_Checks_/_de.unkrig.NameSpelling)

If you like the concept, then you could adapt it for sevntu-checkstyle, and maybe save a lot of effort in the future.

aunkrig:
Hi Roman,

(1) I can't find the link to Lars's thoughts on the issue any more - and your link above does not take me to the right place.

(2) I adopted on Lars's ideas and redesigned the cs doclet such that it is now based on ANNOTATIONS - only the lengthy "rule descriptions" an "property descriptions" are still taken from doc comments. Surprisingly JAVADOC+DOCLET are very powerful when it comes to processing annotations - works like a breeze :-) .

The links above are still valid and now point to the updated documents.

What do you think about it?

romani:
@lkoe, please restore your idea if that possible, looks like idea is about to become implemented.

@aunkrig, thanks for you time, I will come back to you proposal as soon as Eclipse-CS 6.2 will be released, as I sevntu project will require some time to work with new version Eclipse-CS.

lkoe:
I've rebased the branch with my metadata annotations draft (a very rough one at that) to the current cs master:
[...]

PR #327, opened 2015-03-01
Checking if there is no switch block inside other switch block.

damianszczepanik:
Issue #249

### DIFF
--- a/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.properties
+++ b/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.properties
@@ -102,6 +102,9 @@ NameConventionForJunit4TestClassesCheck.expectedClassNameRegex = Regular express
 NameConventionForJunit4TestClassesCheck.classAnnotationNameRegex = Regular expression which matches test annotation names on classes.
 NameConventionForJunit4TestClassesCheck.methodAnnotationNameRegex = Regular expression which matches test annotation names on methods.
 
+NestedSwitchCheck.name = Nested switch should be avoided
+NestedSwitchCheck.desc = Nested switch should be avoided, extract nested switch block into separate method.
+
 NoMainMethodInAbstractClass.name = No Main Method In Abstract Class
 NoMainMethodInAbstractClass.desc = Checks if an abstract class does not have "main()" method, because it can mislead a developer to consider this class as a ready-to-use implementation
 
--- a/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.xml
+++ b/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.xml
@@ -403,6 +403,11 @@
             <message-key key="incorrect.getter.name" />
             <message-key key="incorrect.setter.name" />
         </rule-metadata>
-               
+
+        <rule-metadata name="%NestedSwitchCheck.name" internal-name="NestedSwitchCheck" parent="TreeWalker">
+            <alternative-name internal-name="com.github.sevntu.checkstyle.checks.design.NestedSwitchCheck"/>
+            <description>%NestedSwitchCheck.desc</description>
+        </rule-metadata>
+
     </rule-group-metadata>
 </checkstyle-metadata>
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/NestedSwitchCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/NestedSwitchCheck.java
@@ -1,0 +1,79 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2015  Oliver Burn
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+package com.github.sevntu.checkstyle.checks.design;
+
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.checks.coding.AbstractNestedDepthCheck;
+
+/**
+ * <p>
+ * This check ensures that there is no switch block inside other switch block. In such case nested
+ * block should be exposed into new method.
+ * </p>
+ * Assuming following block:
+ * 
+ * <pre>
+ * <code>
+ *      switch (color) {
+ *      case GREEN:
+ *          switch (type) {
+ *          case MEDIUM:
+ *          }
+ *      case BLUE:
+ *      }
+ * </code>
+ * </pre>
+ * 
+ * Nested switch block that checks <code>type</code> parameter should be converted into separate
+ * method.<br>
+ * To enable this check use following configuration:<br>
+ * <br>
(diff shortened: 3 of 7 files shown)


romani:
0)
please before you start working on issue please let me know to let me assign it to you,  to avoid collisions in implementations.

1) 
http://checkstyle.sourceforge.net/availablechecks.html
>
NestedForDepth 	Restricts nested for blocks to a specified depth (default = 1).
NestedIfDepth 	Restricts nested if-else blocks to a specified depth (default = 1).
NestedTryDepth 	Restricts nested try-catch-finally blocks to a specified depth (default = 1).

I do believe that we need to follow that naming notation.

2)
> avoid.nested.switch=Extract nested switch block into the method.

please before suggesting user what to do, please state a reason of violation.

3) please take a look at Javadocs of checks that are in main library, we need examples of config and examples of code to show what we are going to catch. Recent example: https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/blocks/EmptyCatchBlockCheck.java

4) We need to test your Check on opensource projects, please do report generation to prove that you do not have Exceptions on non standard code, UTs usually catch only 80% of problems.
example: http://andrew0701.github.io/RequiredParameterForAnnotation_issue321/checkstyle.html
https://github.com/Andrew0701/checkstyle-tester
testing over private code is also required.

damianszczepanik:
Updated 1,2 and 3 however I have problem with 4. I launched tester tool it complained about my test (it could not be found) but I've fixed it. However looks like no matter what I put into the xml file I get always the same results that do not contain my rules.

Is this correct configuration file (I want to test only my check):
```
<module name = "Checker">
    <property name="charset" value="UTF-8"/>
    
    <property name="severity" value="warning"/>
    
    <module name="TreeWalker">
        <module name="com.github.sevntu.checkstyle.checks.design.NestedForSwitchCheck"/>
    </module>
</module>
```

and the build output:
```
[INFO] Generating "Checkstyle" report    --- maven-checkstyle-plugin:2.9.1
[INFO]
[WARNING] Unable to locate Source XRef to link to - DISABLED
[INFO] Generating "Checkstyle" report    --- maven-checkstyle-plugin:2.9.1
[INFO]
[WARNING] Unable to locate Source XRef to link to - DISABLED
[INFO] Generating "Dependency Convergence" report    --- maven-project-info-reports-plugin:2.8
[INFO] Generating "Dependency Information" report    --- maven-project-info-reports-plugin:2.8
[INFO] Generating "About" report    --- maven-project-info-reports-plugin:2.8
[INFO] Generating "Plugin Management" report    --- maven-project-info-reports-plugin:2.8
[INFO] Generating "Project Plugins" report    --- maven-project-info-reports-plugin:2.8
[INFO] Generating "Project Summary" report    --- maven-project-info-reports-plugin:2.8
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
```

romani:
please make name of check as NestedSwitchCheck.

I will review your code/problem a bit alter.

romani:
@damianszczepanik , 

> Is this correct configuration file (I want to test only my check):

Pelase review that example file and project, I see difference in naming a Check in config:
https://github.com/sevntu-checkstyle/checkstyle-samples/blob/master/maven-project/checkstyle.xml

let me know if you even after that have problems.

romani:
You missed update at checkstyle-extensions.xml, please refer to another New Check commit https://github.com/sevntu-checkstyle/sevntu.checkstyle/commit/16e829e0f018ed3aa5ac48ff66f408eca658fd56

damianszczepanik:
Ok, it took me a while to make it 'work'. However the tool you provided probably works pretty nice for pure checkstyle checks but won't for sevntu. It does not build the sources (`mvn clean site`) so there is no jar file with my new test from sevntu. It uses checks from .m2 directory so when you `mvn package` in checkstyle project you have access to your tests. Since it does not compile the project there is problem to extend `checkstyle_packages.xml` for sevntu packages.

Finally I moved my test to checkstyle project to com.puppycrawl and then it started work. I will publish my results later and maybe in meantime I update this tool so it works more friendly also with new checks from sevntu...

First question would be: can I configure `checkstyle_packages.xml` from `pom.xml` like this one http://maven.apache.org/plugins/maven-checkstyle-plugin/examples/custom-checker-config.html ?

romani:
I do not remember that it is possible.

Sorry that checkstyle and its extension is hard to use, we will definety resolve that. It should be simple.

Please send me link to your code ( branch), i will provide you a solution in a day or two.

romani:
I updated that checkstyle-tester and put it for hosting at https://github.com/checkstyle/contribution/tree/master/checkstyle-tester , as this project become mature we could move it in separate project.

What I did : 
1) git clone https://github.com/damianszczepanik/sevntu.checkstyle.git
2) cd checkstyle
3) git checkout AvoidNestedSwitchCheck 
4) ./deploy-all.sh 

all jar are appear in local ".m2"

5) git clone https://github.com/checkstyle/contribution.git
6) cd checkstyle-tester
7) activate your Check in  my_check.xml 
8) ./launch.sh -Dcheckstyle.config.location=my_checks.xml
9) see result HTML report at target/site/index.html

Oh my god !!! checkstyle code do has nested switch :( .

damianszczepanik:
I found this version of checkstyle-tester more friendly however I found some errors that make this version hard to run on clean repo :) (https://github.com/checkstyle/contribution/pull/2).

And here we have report (http://damianszczepanik.github.io/sevntu.checkstyle/sevntu.checkstyle/pull/327/index.html) and a few questions about:
1. TreeWalker reports error `Got an exception - unexpected token: Session` which I guess we should worry about but this is not fault of my new check.
2. My check is not reported in Rules section. If I could guess this is because it is not from checkstyle package/repo. Would be nice if you could fix if some day
3. I've uploaded only summary so let me know if there is anything else you need to make this PR completed.

romani:
0) thanks a lot for contribution 

1) As you can see from your report you launched tester against AbstractClassName Check. Please redo report generation for your Check.
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/my_check.xml#L17
Please look at few lines above for your Check.
I put AbstractClassName as example only.

2) 
> apache-ant/src/tests/antunit/taskdefs/javac-dir/bad-src/Bad.java
 hibernate-orm/documentation/src/main/docbook/devguide/en-US/chapters/events/extras/JpaCallbacksExample.java
....

It is a bad sources (non compilable), we do not support non compilable sources. so it is OK.
We just need to find way to skip them from testing.

damianszczepanik:
Updates check generation http://damianszczepanik.github.io/sevntu.checkstyle/sevntu.checkstyle/pull/327/index.html but still the Rules section is empty.

romani:
It could be empty when rule is used with default parameres. i think that should be fixed in maven plugin

romani:
Issue MCHECKSTYLE-289 - enableRulesSummary option does not show Rules with default configuration has been successfully created (http://jira.codehaus.org/browse/MCHECKSTYLE-289).

romani:
I do not see references to sources , please submit them to github , updated version https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/README.md

damianszczepanik:
Adding to commit 26k files / 05GB takes minutes, uploading even longer and my Git client crashed when executing this operation! :) Does it make sense to upload whole target/site directory? I expect you want to have no crash and if so you rather won't to check sources otherwise you want to fix the crash and also won't to look at source code?

push updated

romani:
please split amount of project in few groups or take the most representative projects.
I would stay with maven projects only to let checkstyle skip validation base on folder pattern and avoid testing over UT, resources and other non related to compilable java clases .
We need any resonable way to share report(s).

damianszczepanik:
The first run should be done against as many projects as possible to check whether new check crashes or not. Second should be done with smaller number of projects to prove that check is valid.

Updated http://damianszczepanik.github.io/sevntu.checkstyle/sevntu.checkstyle/pull/327/index.html with those sources that cause violations.

romani:
> The first run should be done against as many projects as possible to check whether new check crashes or not. Second should be done with smaller number of projects to prove that check is valid.

Good words!. Please contribute them for future generation to README file.

> those sources that cause violations

Good report, I am merging your Check, thanks a lot!

romani:
stop .....

Not all files are updated, please review that commit as example of files that need to be updated - https://github.com/sevntu-checkstyle/sevntu.checkstyle/commit/16e829e0f018ed3aa5ac48ff66f408eca658fd56 , eclispe-cs, sonar, ......

damianszczepanik:
It is not working the way I like. That would be cool if we have tool that will verify xml and properties files when `mvn verify` so I don't need to remember which files shall be updated, no need to copy the same descriptions in different places and maintain both in case of typo.

Any idea for this problem?

romani:
Yes ,it is pain already for few years, but cost(time to spend) of that tool is bigger than pain to do it manullay, so every one step on that problem and desided to make it manually again and again.

FYI: https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/290

All contribution or ideas for this issue are welcome.

damianszczepanik:
Any update about this one?

romani:
I lost an event that you updated a code and just waited a reply from you, sorry, merged as FF.
thanks for pinning me.

PR #329, opened 2015-03-07
Remove unused input for AvoidHidingCauseExceptionCheck

damianszczepanik:


damianszczepanik:
Removed file that is not used for testing (probably never was ). I confirmed that cases provided by this file were already covered by two other test classes.

romani:
merged as FF, thanks a lot !

PR #370, opened 2015-08-04
Implemented UniformEnumConstantNameCheck

pbaranchikov:
Removed logic to distinguish enum objects from enum constants.
Added support for multiple format of enum values. Enum is allowed if there is at least one format, that hist values conform.
Default formats are both CamelCase and UPPER_CASE

Check for guava

formats value | Number of violations found
------------- | -------------
default | 0
"^[A-Z][a-zA-Z0-9]*$" (CamelCase) | 59
"^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$" (UPPER_CASE) | 12

Reports:
http://pbaranchikov.github.io/checkstyle/enum-value-uppercase/checkstyle.html
http://pbaranchikov.github.io/checkstyle/enum-value-camel/checkstyle.html

### DIFF
--- a/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/naming/checkstyle-metadata.properties
+++ b/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/naming/checkstyle-metadata.properties
@@ -7,3 +7,7 @@ EnumValueName.desc = <p>Check forces enum values to match the specific pattern. 
 EnumValueName.constFormat = Regex for Values Enumeration names to conform
 EnumValueName.objFormat = Regex for Class Enumeration names to conform
 EnumValueName.excludes = Exclude method and field names regexp list (comma-separated)
+
+UniformEnumConstantNameCheck.name = Enum values name
+UniformEnumConstantNameCheck.desc = Check forces enum values to match one of the specified patterns and forces all the values to follow only one of the specified patterns.
+UniformEnumConstantNameCheck.formats = Regex list that enum value names should conform
--- a/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/naming/checkstyle-metadata.xml
+++ b/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/naming/checkstyle-metadata.xml
@@ -25,5 +25,13 @@
                 <description>%EnumValueName.excludes</description>
             </property-metadata>
         </rule-metadata>
+
+        <rule-metadata name="%UniformEnumConstantNameCheck.name" internal-name="UniformEnumConstantNameCheck" parent="TreeWalker">
+            <alternative-name internal-name="com.github.sevntu.checkstyle.checks.naming.UniformEnumConstantNameCheck" />
+            <description>%UniformEnumConstantNameCheck.desc</description>
+            <property-metadata name="formats" datatype="String" default-value="^[A-Z][a-zA-Z0-9]*$,^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$">
+                <description>%UniformEnumConstantNameCheck.formats</description>
+             </property-metadata>
+        </rule-metadata>
     </rule-group-metadata>
 </checkstyle-metadata>
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/naming/UniformEnumConstantNameCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/naming/UniformEnumConstantNameCheck.java
@@ -1,0 +1,229 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2016 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.sevntu.checkstyle.checks.naming;
+
+import java.util.ArrayList;
+import java.util.BitSet;
+import java.util.Collection;
+import java.util.List;
+import java.util.regex.Pattern;
+
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+/**
+ * Check forces enum constants to match one of the specified patterns and forces
+ * all the values to follow only one of the specified patterns.
+ *
+ * <p>By default both CamelCase and UPPER_CASE are allowed, so check validates,
+ * whether all the values conform the either of them.
+ *
+ * <p>For example, both enums are allowed by the check:<pre>
+ * public enum EnumOne {
+ *    FirstElement, SecondElement, ThirdElement;
+ * }
+ * public enum EnumTwo {
+ *    FIRST_ELEMENT, SECOND_ELEMENT, THIRD_ELEMENT;
+ * }</pre> But the following enum, is violated, because values conform
+ * different notations: <pre>
+ * public enum EnumThree {
+ *    FirstElement, SECOND_ELEMENT, ThirdElement;
+ * }
+ * </pre>
+ *
+ * <p>To use only CamelCase, use the following configuration:
+ *
+ * <pre>
+ * &lt;module name="UniformEnumConstantNameCheck"&gt;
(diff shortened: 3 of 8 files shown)


romani:
postponed to 1.14 as it is braking compatibility changes

romani:
Sorry for long delay in reply .

1) 
>```  * By default both CamelCase and UPPER_CASE are set, so check
 * <ul>
 * <li>checks for all the values conform one of the specified patterns
 * <li>checks for all the values conform the same pattern.
```

I do not understand this description, please try rephrase. Two point are conflicting.

2)
> But the following enum, is rejected, because elements conform

elements or values ? lets be consistent , please review JDK specification for correct name.
rejected ==> violated

3)
> To change the defaults, use {@code formats} property

Check configuration is not trivial, please show example. Users always demand more examples of configuration in Checkstyle.

4) 
> Only first violation is reported for each enumeration.

Please remind me why we do this.

5) 
> ```
    public final void setFormats(String[] regexps)
    {
        patterns = new ArrayList<Pattern>(regexps.length);
```

Please provide example of two regexp format that are separated by ",". In config we have only String type for option values.

6)
> public int[] getDefaultTokens()

please override getRequiredTokens and getAcceptableTokens

7) 
> for (DetailAST member = objBlock.getFirstChild(); member != null; member = member

please wrap line by ";"

8) 
>```
final Collection<Pattern> initialPatterns = new ArrayList<Pattern>(patterns);
            for (final Iterator<Pattern> it = patterns.iterator(); it.hasNext();) {
                final Pattern pattern = it.next();
                if (!pattern.matcher(enumName).find()) {
                    it.remove();
                }
            }
```

we can not change option values base on some data from current java file. Check object is reused between all input java files - options should be untouched. Somewhere in future we will enforce that immutability by design, sorry that you run into design problem of Checkstyle.

9)
> ```
for (DetailAST member = objBlock.getFirstChild(); member != null; member = member
                .getNextSibling()) {
            if (member.getType() != TokenTypes.ENUM_CONSTANT_DEF) {
```

please review out Utils , there should be method like getNextSiblingToken(TokenTypes type) or similar.
Formats !=== patterns , this is not critical but not desirable.

romani:
@pbaranchikov , can we finish this PR ?

romani:
@pbaranchikov , do you plan to finish this PR ?

If you have no time, just let me know, I find another contributor to finish.

pbaranchikov:
1. done
2. done
3. done
4. There is no mean to report all the values, that has been violated, it's sufficient to report the first one. But internal logic make it rather hard to report all the violations, for example, if you allow both UPPER_CASE and CamelCase, what certain values should be reported, if they are
``` java
FirstValue, SECOND_VALUE, ThirdValue, FORTH_VALUE
```
5. done in javadoc
6. done
7. done
8. I have not modified pattern field, there is a string earlier:
``` java
        final Collection<Pattern> patterns = new LinkedList<Pattern>(this.patterns);
```
So, to avoid the confusion, I've renamed local variable "patterns" to "conformedPatterns"
9. I had not found anything suitable in com.github.sevntu.checkstyle.Utils class. There are only 2 methods there. Did you mean some other Utils class?

romani:
1)
> sevntu-checks/.classpath

please explain changes in this file. I do not expect this file to be changed.

2)
> 
```
        for (DetailAST member = objBlock.getFirstChild(); member != null;
                member = member.getNextSibling()) {
            if (member.getType() != TokenTypes.ENUM_CONSTANT_DEF) {
                continue;
            }
```

this method should looks better:
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/DetailAST.java#L129

2)
>             final String enumName = member.findFirstToken(TokenTypes.IDENT)
                    .getText();

please make single line.

3)
> public final void setFormats(String[] regexps)

this check (with such name) should simply test names by regexp as all other Checks do this http://checkstyle.sourceforge.net/config_naming.html.

Your idea to make a rule to validate names to conform certain conventions but only one convention in one Enum - is valid. But this is just another Check.

pbaranchikov:
1. done (changes added accidentally)
2. I still do not see the solution you are pointing to. I need to go though all the children of the specified type. How can `addNextSibling(DetailAST ast)` be usefull?
2. Fixed
3. Lets imagine, that I'll implement a separate check to validate whether enum value name _conform certain conventions but only one convention in one Enum_.
In this case, the check will fully cover the EnumValueNameCheck, because checking for one pattern is a special case of checking for multiple patterns.

romani:
> How can addNextSibling(DetailAST ast) be usefull?

my bad, I thought that we already created such util function as `getNextSibling(TokenType)`.

> In this case, the check will fully cover the EnumValueNameCheck, because checking for one pattern is a special case of checking for multiple patterns

My main problem to this solution is difference in design with other Naming Checks - http://checkstyle.sourceforge.net/config_naming.html#MemberName. They do validate names independently , but your request enforcing dependency between them.

Yes overlapping, but user will have simple way to enforce naming in Enums. EnumValueNameCheck will be first candidate to go to main project to enforce no "$" in names. Minor benefit but still benefit.  ("_" is in demand by default https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names). But some people do not like upper-case values at all, or to opposite like to keep all enum to be capital.

More advanced naming convention Check (consistency in format) will stay a bit longer to let test in real code and examine more Enum cases, idea is great but it is advanced management over names (like difference between http://checkstyle.sourceforge.net/config_naming.html#MemberName and http://checkstyle.sourceforge.net/config_naming.html#AbbreviationAsWordInName).

By such new check you can enforce no "_" if names are started from lowercase (mostlikely camelcased and do not need "_"). But Enums are used a lot to have identifier in java and its toString value(used in generation of sql or URL ) the same - format could be any.
 
Overlapping of functionality is ok, we already have several examples of these - user will choose himself what Check to use.

Enum values are could be treated as "public fields" of class - class members. Why do not we have list of formats for http://checkstyle.sourceforge.net/config_naming.html#MemberName , kind of the same case could happen here too.

----------

If you have more evidences please share them, I need to know all reasons before I allow to introduce Check that is sitting in a family of Checks but it different by principle of validation.

pbaranchikov:
According to conversation, renamed this check to a new check - UniformEnumValueName

romani:
@pbaranchikov ,

sorry for delay in reply.

1)
> 
```
    public int[] getRequiredTokens()
    {
        return ArrayUtils.EMPTY_INT_ARRAY;
    }
```

What is a reason in `ArrayUtils.EMPTY_INT_ARRAY` ? I think it should be like `getAcceptableTokens()`, and please call `getAcceptableTokens()` from `getDefaultTokens()`


2)
> public void visitToken(DetailAST ast)

Logic of this Check is very simple but code looks complicated and hard to read.
Please make code like or even better if you see other way:
```
final DetailAST objBlock = ast.findFirstToken(TokenTypes.OBJBLOCK);
List<DetailAST> members = getChildNodesOf(TokenTypes.ENUM_CONSTANT_DEF)
for (DetailAST member : members) {
    boolean multipleMatches = match(member.getText(), this.patterns);
    if (multipleMatches) {
       .....
    }
}

static private boolean match(....) {
 //no copy of patterns, please create Bitset collection to mark index of matched pattern and use its field that counts its bits.
}
```

3) 
3 copy of "this.patterns" looks strange for me. Most likely will be addressed in "2)".

4)
> `<description>%UniformEnumValueName.desc</description>^M`

not sure about '^M', do we need it ?

5)
> `if (initialPatterns.size() == 1) {`

can we make it a as field that is assigned at setter `public final void setFormats(String[] regexps)`  ?

6)
Please provide regression testing results for opensource projects, please use checkstyle-tester (https://github.com/checkstyle/contribution/tree/master/checkstyle-tester) tool to generate report.

pbaranchikov:
1. done
2. done
3. done
4. done
5. done
6. done new links are: http://pbaranchikov.github.io/checkstyle/uniform-enum-value-default/checkstyle.html
and http://pbaranchikov.github.io/checkstyle/uniform-enum-value-uppercase/checkstyle.html
7. done

From #445 :
0. done
1. done
2. done

romani:
@pbaranchikov ,

8)
your:
```
        for (DetailAST member = objBlock.getFirstChild(); member != null
                && !conformedPatterns.isEmpty();
                member = member.getNextSibling()) {
            if (member.getType() != TokenTypes.ENUM_CONSTANT_DEF) {
                continue;
            }
```
vs mentioned by me
```
List<DetailAST> members = getChildNodesOf(TokenTypes.ENUM_CONSTANT_DEF)
for (DetailAST member : members) {
```

Why do you need "&& !conformedPatterns.isEmpty();" ? (looks like it could limit validation of all children)
Please add UTs case where violation is second and last element of enum.

pbaranchikov:
Due to the check idea, second and last constant cannot be invalid. Only the first invalid enum constant is reported. This is because we originally do not know, what exact pattern user used. Consider the example:
```java
public enum EnumExample {
    first, SECOND, THIRD;
}
```
It seem to be incorrect to report, that _SECOND_ and _THIRD_ are invalid, because it may be a legal case, when _first_ should be fixed instead. That's why it is pointless to validate every constant. We should validate the whole enum instead.

romani:
ok, please update javadoc to explain this. 
This is not obvious.

romani:
@pbaranchikov ,

there is a build failure due to wrong javadoc - https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/jobs/126612442#L1176
Please fix.

pbaranchikov:
@romani, could you please help with resolving the issue. According to build results, one of the JavaDoc violations is only triggered, when executing from openjdk8 - https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/builds/131340872. It looks suspicious.

romani:
@pbaranchikov ,

problem is stable, the same on my local. Just run "mvn javadoc:javadoc" (it is located at .travis file) and resolve the issue.

here is a commit that fixed all similar violations - https://github.com/sevntu-checkstyle/sevntu.checkstyle/commit/8f4ac566efe1e697207e8f202993b715cb5ed109

romani:
thanks a lot for fix !!!
merged.

Issue #355, opened 2015-06-10
CauseParameterInExceptionCheck: NullPointerException occurred during the analysis


rdiachenko:
```
public enum A{
	B;
	private A() {}
}
```

```
<module name="CauseParameterInException">
      <property name="severity" value="warning"/>
</module>
```

```
[ERROR] NullPointerException occurred during the analysis of file /SomeClass.java.
java.lang.NullPointerException
        at com.github.sevntu.checkstyle.checks.design.CauseParameterInExceptionCheck.getClassDef(CauseParameterInExceptionCheck.java:244)
        at com.github.sevntu.checkstyle.checks.design.CauseParameterInExceptionCheck.visitToken(CauseParameterInExceptionCheck.java:154)
        at com.puppycrawl.tools.checkstyle.TreeWalker.notifyVisit(TreeWalker.java:393)
        at com.puppycrawl.tools.checkstyle.TreeWalker.processIter(TreeWalker.java:482)
        at com.puppycrawl.tools.checkstyle.TreeWalker.walk(TreeWalker.java:322)
        at com.puppycrawl.tools.checkstyle.TreeWalker.processFiltered(TreeWalker.java:195)
        at com.puppycrawl.tools.checkstyle.api.AbstractFileSetCheck.process(AbstractFileSetCheck.java:76)
        at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:260)
        at org.apache.maven.plugin.checkstyle.exec.DefaultCheckstyleExecutor.executeCheckstyle(DefaultCheckstyleExecutor.java:253)
        at org.apache.maven.plugin.checkstyle.AbstractCheckstyleReport.executeReport(AbstractCheckstyleReport.java:473)
        at org.apache.maven.plugin.checkstyle.CheckstyleReport.executeReport(CheckstyleReport.java:155)
        at org.apache.maven.reporting.AbstractMavenReport.generate(AbstractMavenReport.java:255)
        at org.apache.maven.reporting.AbstractMavenReport.execute(AbstractMavenReport.java:143)
        at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:134)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:208)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
[...]

PR #380, opened 2015-10-07
Fix NPE in CauseParameterInExceptionCheck

igieon:
Fix issue #355


romani:
Thanks a lot !

merged as FF, without version bump.

will be released as somewhere next 24 hours.

PR #384, opened 2015-10-25
Issue#334 Add new check SingleSpace

robertwhitebit:


mkordas:
@robertwhitebit 
Thank you for really great check. I've enforced it over my commercial project without any false-positives or exceptions. I think we should consider moving it into Checkstyle repo.

Issue #401, opened 2015-11-11
Sonar plugin does not work with sonar-checkstyle-plugin v2.4


cypai:
I get a guava dependency error during project analysis with the most recently released version of sonar-checkstyle-plugin (v2.4).

The same issue is mentioned here:
https://groups.google.com/d/msg/sonarqube/u9ZLuLc0ZXc/aPGhBA1LBwAJ

Related issue:
https://groups.google.com/d/msg/sonarqube/we1uaK570m8/QAukuq9Mij0J

The sonar-checkstyle-plugin developers were forced to shade guava into their jar because of compatibility issues between SonarQube's old version of guava and the main checkstyle project. Their suggested fix for sevntu is to shade guava into the sevntu-checkstyle-sonar-plugin jar like they did. I can confirm that shading the jars will fix both of the issues linked above.

Link to their shading pom: https://github.com/SonarCommunity/sonar-checkstyle/blob/4a01ebcbd1ba176f0975e04b31c3312b78ff2c43/checkstyle-all/pom.xml

Error stack trace here:
[ERROR] Failed to execute goal org.codehaus.mojo:sonar-maven-plugin:2.6:sonar (default-cli) on project stuff: com.puppycrawl.tools.checkstyle.api.FileContents.getCComments()Lcom/google/common/collect/ImmutableMap; -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:sonar-maven-plugin:2.6:sonar (default-cli) on project stuff: com.puppycrawl.tools.checkstyle.api.FileContents.getCComments()Lcom/google/common/collect/ImmutableMap;
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:216)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:80)
	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:51)
[...]

PR #406, opened 2015-11-17
Issue #401: Shade Guava in sonar-plugin

Vladlis:


romani:
please provide evidences that it works in Sonar, some screen-shots is ok.

Vladlis:
@romani
default

romani:
merged as FF

Issue #417, opened 2015-12-28
Rename checks to follow Checkstyle convention for check names


MEZk:
https://github.com/checkstyle/checkstyle/issues/2780

Checkstyle name convension for checks names: ```XXXXCheck```

The following classes should be renamed:
AvoidDefaultSerializableInInnerClasses
ForbidCCommentsInMethods
ReturnBooleanFromTernary
ReturnNullInsteadOfBoolean

romani:
fix is merged.

PR #418, opened 2015-12-28
Issue #417: Rename checks to follow Checkstyle convention for check names

MEZk:


romani:
such names should be all around configuration files, not only in code.

MEZk:
@romani 
Why we should use the whole name in xml? In main project we use names without ```Check```.

romani:
I do not remember, but there are bunch of configs for different tools, some of them use their own way of Check loadings from class path.

MEZk:
@romani 
Corrected.

romani:
merged as FF

Issue #374, opened 2015-09-13
New check: only one break or continue statement should be inside cycle body


sabaka:
Restricting the number of break and continue statements in a loop is done in the interest of good structured programming.

One break and continue statement is acceptable in a loop, since it facilitates optimal coding. If there is more than one, the code should be refactored to increase readability.

inspired by: http://nemo.sonarqube.org/coding_rules#languages=java|q=one%20break

that rule already enforced over Checkstyle codebase.

yaziza:
Hi, i wanna contribute to the project and i think this issue can be a good starting  as it s labeled easy.

Can you please give some hint where to start ?
Thank you !

romani:
new contributors are always welcome.

please read https://github.com/sevntu-checkstyle/sevntu.checkstyle/wiki
http://checkstyle.sourceforge.net/extending.html
http://checkstyle.sourceforge.net/contributing.html
http://checkstyle.sourceforge.net/beginning_development.html

please join this mail-list https://groups.google.com/forum/#!forum/sevntu-checkstyle , open topic , an there will discuss all your questions and problems.

yaziza:
Hi, i have started working on this issue, i will fix all checkstyle violations and write unit tests on the next few days. Feedback highly appreciated :)

romani:
please finish with checkstyle violations first and then share a link for review.

romani:
merged.

Issue #410, opened 2015-12-19
AvoidModifiersForTypes throws Exception while processing modifier token which has type LITERAL_PRIVATE 


MEZk:
AvoidModifiersForTypes throws Exception while processing modifier token which has type LITERAL_PRIVATE. The problem is appeared due to the fact that the check does not take into account LITERAL_PRIVATE modifier of the variable.

See https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/81a2cf91fbb1d2ec18ffd6f5cfc2bb67828c4834/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidModifiersForTypesCheck.java#L156 

**Configuration:**
```xml
<?xml version="1.0" ?>

<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
        "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">

<module name="Checker">
    <module name="TreeWalker">
        <module name="AvoidModifiersForTypes"/>
    </module>
</module>
```

**Input file:** https://github.com/checkstyle/checkstyle/blob/e8475b41261f724128c869b4de371713d86228c3/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java

SevNTU Checkstyle 1.17.1 output:

```
Exception was thrown while processing checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java: Found unsupported token: LITERAL_PRIVATE 
```

romani:
fix is merged.

romani:
@yaziza , your update is released, please enforce it over checkstyle code - https://github.com/checkstyle/checkstyle/issues/3064

PR #423, opened 2016-01-11
Issue #374: Fix SingleBreakOrContinue violations

yaziza:


yaziza:
Update: 6/7 violations resolved.

Im still working on the last violation. After fixing i will extend the unit tests to meet the requirements.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/CustomDeclarationOrderCheck.java
```diff
@@ -596,37 +596,27 @@ private int getPositionInOrderDeclaration(final DetailAST ast)
...
-                }
-                else if (currentRule.hasRule(MAIN_METHOD_MACRO)) {
-                    if (isMainMethod(ast)) {
+
+                    if (isAnonymousClassField(ast)
+                            || classDetail.containsGetter(methodName)
+                            || classDetail.containsSetter(methodName)
+                            || isMainMethod(ast)) {
                         result = index;
                         break;
```
yaziza: After reducing the number of breaks to 2, not sure how to eleminate the last one ...
Maybe completly refactoring this method ...

romani:
please remove from you commit all unrelated changes(that are just formatting), or move them in separate commit of this PR. This will ease diff verification.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/CustomDeclarationOrderCheck.java
romani: code is not simple :), this is just good example that need to follow new Checks ourself.

solution is to put `result != null` condition to FOR, and remove `break` completely.

romani:
Travis is not happy with reduced coverage
> [ERROR] Jan 11, 2016 7:59:02 PM net.sourceforge.cobertura.coveragedata.CoverageDataFileHandler loadCoverageData
INFO: Cobertura: Loaded information on 63 classes.
com.github.sevntu.checkstyle.checks.coding.DiamondOperatorForVariableDefinitionCheck failed check. Branch coverage rate of 85.0% is below 90.0%

It might be resulted by removing the code lines. Please generate coverage report and share pictures of coverage before and after for this method. If it is the same it is ok to change configuration(see cobertura section `pom.xml`) there limit is specified.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/CustomDeclarationOrderCheck.java
yaziza: Sorry, i didn't get it. 

What about setting: index = Integer.MAX_VALUE - 1; instead f break ?

romani: make loop like `for (int index = 0; index < customOrderDeclaration.size() && result != 1 ; index++) `
no breaks are required.

romani:
@yaziza , 1.18 release will be this Sunday, please find time to finish this PR.

yaziza:
Fixed, i have added a new case in the test input file and the branch cover is now 94%
Also i have removed the unnesseray != null check, which will be checked anyway in equals.

Thank you for your support !

Do you have any suggestions on issues i can tackle next ?

romani:
merged as FF. Thanks a lot for your time and contributions.

> Also i have removed the unnesseray != null check

One of the banefit of 100%  coverage is to remove dead/extra code. 

> Do you have any suggestions on issues i can tackle next ?

Can you help us with #410 ? Exceptions during execution usually force user to deactivate the Check at all. But you can choose any unassigned issue.

Issue #286, opened 2014-12-16
new Check: NonAcessibleType


romani:
Return Logger out of the class is the very bad practice as it produce logs that are hard to investigate as logging class does not contains that code and search should be done in other classes or in hierarchy (if filed is public or accessible by other protected or package)

```
package com.test;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class Check {

    private Logger log = LoggerFactory.getLogger(getClass());

    protected Logger log2 = LoggerFactory.getLogger(getClass()); //Violation

    public Logger log3 = LoggerFactory.getLogger(getClass()); //Violation

    Logger log3 = LoggerFactory.getLogger(getClass()); //Violation

    public Logger getLog() { //Violation
        return log;
    }

}
```

attention: imports needto be considered to reduce amountof false positives.
options: user need to setup regExp for type tio restict for public

The same should be checked for Interfaces and Enums, methods with @Override annotation should be ignored.

PR #426, opened 2016-01-18
Issue #410: Fix exception while processing types with LITERAL_PRIVATE …

yaziza:
…modifiers


yaziza:
seems too easy, not sure if i missed something !

romani:
please put "#" before issue number, to allow github parse it as link.

romani:
and here we have a problem

from Javadoc: 

> Only 4 types according to Java Spec: static, final, transient, volatile. 

but https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.3.1 , so 

we need to extend code to support all modifiers. Javadoc need to be updated to have link to spec and list of all modifiers.

romani:
0) each time you do a fix please do a comment, I expect notification from github. Only code push will not give any notification and I will wait for you.

1)
> ` Initial Implementation`

please make Author tag simple, just append yourself to authors, no details are required.

2)
please add all new options to the javadoc, Javadoc is all that we have for documentation of Check.

3)
put example to javadoc of forbidding `@Autowired` for fields. 

4)
put example of forbidding making public `Log` types. kind of simple implementation of https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/286 .

yaziza:
Patch updated !

Do we need to test all this cases, or just put them in the documentations ?

romani:
0)
 please keep answering on each my point as "Done", that do help track of what is really done and what should be discussed. 

1)
> Example 3

configuration is missed , please add.

2)
> Do we need to test all this cases, or just put them in the documentations ?

yes we need, it is such a shame when official example does not work as described. Testing is main part of development. If any false positive happen, user can deactivate Check completely. So the whole reason of Check update is missed.

Please remove this line and make test coverage 100%
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/pom.xml#L118. To make sure we did not miss obvious cases.

3) 
Lets run configs that I asked you to put in javadoc on a real code to make sure there is no obvious false-positives or exceptions (checkstyle-tester).

yaziza:
1) done

2) and for a good reason, just realised that i forgot the case where the 'protected' is not explicitly written.

But i have trouble testing the default case, is it ok to put non compiling code the test/resource to test this case ? 

3) done: report for checkstyle and sevntu report only the 3 cases i wrote in the input test file:

http://yaziza.github.io/checkstyle.html

Config i used : 

```
    <module name="TreeWalker">
         <module name="AvoidModifiersForTypesCheck">
             <property name="forbiddenClassesRegexpProtected" value="Logger"/>
             <property name="forbiddenClassesRegexpPublic" value="Logger"/>
         </module>
    </module>
```

romani:
1)
> But i have trouble testing the default case, is it ok to put non compiling code the test/resource to test this case ? 

No, all inputs have to be compilable.

2)
>  Warning 	extension 	AvoidModifiersForTypes 	'Logger' instance should not have 'protected' modifier. 	37

that is not a "protected", this is "package-private" https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html.
This is no-modifier case. Add one more option forbiddenClassesRegexpPackagePrivate.

3)
Please generate report over non-checkstyle sources to catch smth new.

4)
> Example 1: 
Example 2:
Example 3: 
 
Please put in javadoc xml configuration block of Check module to let user copy-paste it to his config and make it work.

5)
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/pom.xml#L118

please remove this line.

yaziza:
1) ok

2) done

3) done : (FindBugs, Guava)

http://yaziza.github.io/checkstyle.html

will fix all violations and send PR

4) done

5) done, but now i have trouble testing the default case, since we only have 8 different allowed modifiers?

Thank you !

romani:
@yaziza , CI need to be green.
https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/jobs/105279714#L807

> 5) done, but now i have trouble testing the default case, since we only have 8 different allowed modifiers?

You will never reproduce this on compilable sources and correctly worked parser.
BUT 100% coverage is so valuable that any hack to archive it make sense.

>  Set<Integer> modifiersSet = getModifiers(ast);

You need to put 'modifiersSet' such unexpected TokenType to go to "default:" .
So you need to modify `ast`.
So you need manually call visitToken method from UTs https://github.com/yaziza/sevntu.checkstyle/blob/e494528ebc6ff04ff345e67c62138fbef97ceeab/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidModifiersForTypesCheck.java#L339 and make `ast` yourself to contain unexpected TokenType.

In UTs please put a comment why you do such a HACK. It is ok.

yaziza:
Hi @romani, sorry i was busy last week, i will finish this patch tomorrow !

romani:
@yaziza , let me know if you need help.

romani:
@yaziza , ping.

yaziza:
Hi @romani, 

sorry i was crazy busy the last few weeks with restricted internet connection ...
I haved added a test case with an unexcpected TokenType, but the testcase was not invoked by Junit ?

Do you know maybe why ? is there any restrictions to consider when extending BaseCheckTestSupport ?

Thank you !

romani:
1)
> but the testcase was not invoked by Junit ?

please point me to exact line. Coverage is  100% now, so I do not understand what you try to archive.


2)
> 
```
                    case TokenTypes.ANNOTATION:
                        if (forbiddenClassesRegexpAnnotation.matcher(className)
                                .matches())
                        {
                            log(ast, MSG_KEY, className, "annotation");
                        }
                        break;
```

that is too obvious copy-paste, please squash all cases in one and make a function to convert TokenType to string.

to be looks like:
```
case:
case:
case:
 if (match(modifierType, className)) {
    log(ast, MSG_KEY, toString(modifierType))
}
```

romani:
@yaziza , ping.

yaziza:
Hi @romani 

1) done, eclipse was just ignoring that test.

2) sorry, didn't get it.  

```
case:
case:
case:
 if (match(modifierType, className)) {
    log(ast, MSG_KEY, toString(modifierType))
}
```

Even if we squash all cases in one, we still need a kind of mapping from TokenType  to patterns (forbiddenClassesRegexpProtected, forbiddenClassesRegexpPublic etc.) And what about the default case ?

```
                    default:
                        Utils.reportInvalidToken(modifierType);

```

romani:
> we still need a kind of mapping from TokenType to patterns 

yes, performance wise it is not good, but it will be so friendly for human to read. We are more focused on human friendly code rather than performance.

> And what about the default case ?

keep it as is. No need to do anything.

yaziza:
Hi @romani ,

patch updated, I hope I didn't misunderstand you ! 

Any suggestion what issue to tackle next ?

romani:
thanks for update , I expect code like:

```
private boolean match(int modifierType, String className) {
        Pattern pattern = mapToRegExp(modifierType);
        boolean match = matcher(className).matches();
}

...

private Pattern mapToRegExp(..... ) {
          switch (modifierType) {
            case TokenTypes.ANNOTATION:
               result = forbiddenClassesRegexpAnnotation;
               break;
            case TokenTypes.FINAL:
                result = forbiddenClassesRegexpFinal;
                break;
              .....
          return result;
}
```

romani:
as new fields are introduced we need to update metadata files - https://github.com/sevntu-checkstyle/sevntu.checkstyle/wiki/List-of-files-that-need-to-be-updated-for-new-field-or-new-Check-implementation

romani:
> Any suggestion what issue to tackle next ?

You can choose yourself, but it will be good to recheck https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/432

yaziza:
Hi @romani ,

Oh i see, i will update my patch today and update the metadata today!

There was issue with Java 8 Build (Report submission to Coveralls API failed with HTTP status 422).

Thanx for your guidance :)

yaziza:
done !

romani:
good, please update all meta data files and I will merge this fix. I expect 2 PRs.

romani:
merged as FF, thanks a lot for all your time and effort to make it happen

Issue #412, opened 2015-12-21
New check: ConstructorWithoutParamsCheck to catch Throwable constructors usage with no arguments


zerg-junior:
An exception message  must contain enough info to find out why an exception occurred (e.g. "Effective Java", item 63). So a check that forbids 
* using Throwable no-args constructors 
* passing empty String literals to Throwable constructors.

may be worth considering.

romani:
first request should be done as general new Check - NoDefaultCtorUsage or smth similar
second request should be as EmptyExceptionMessage, user define regexp of how to distinguish classes from Exception/Error/Throwable

zerg-junior:
@romani  I consider writing the `NoDefaultCtorUsage` check. Is anyone already working on it ?

> second request should be as EmptyExceptionMessage, user define regexp

This is the project-specific `RegExp` and not a new check, right ?

romani:
@zerg-junior ,

> I consider writing the NoDefaultCtorUsage check. Is anyone already working on it ?

I try to keep assignments of issue (by comment or Assigneed field) if somebody what to implement issue.
Nobody is focused on this.
We do not have a constantly focused engineers on the project, you are always welcome to help with implementations

> This is the project-specific RegExp and not a new check, right ?

customized RegExp Check will not help to cover case reliably without false positives.
Special new Check will be more reliable but still not 100% working solution. Checkstyle have limitation and do not have information about Types and their hierarchy - we just have tree of strings.
http://checkstyle.sourceforge.net/writingchecks.html#Limitations
So wee need to help Check to catch Exception types by RegExp like ".*Exception|.*Error|.*Throwable" , if we find code like "new MyException("");" -- you raise an violation.
I could provide you more details and guide you through implementation, just let me know if you with to continue.

zerg-junior:
> you are always welcome to help with implementations

Great, then please skim the following contract:

#### NoCtorWithoutParamsCheck

Prohibits usage of parameterless constructors, including the default ones.

Rationale: constructors of certain classes must always take arguments to properly instantiate objects. Exception classes are the primary example: their objects must  always contain enough info to find out why an exception occurred (see "Effective Java", item 63). Constructing an exception without a cause exception or an exception message leaves out such info and thus should be prohibited.

Configuration: via the `format` field of the `RegExp` type:
```xml
<module name="NoCtorWithoutParamsCheck">
  <property name="format" value=".*Exception|Clazz1"/>
</module>
```


###### Examples
[...]

PR #428, opened 2016-02-16
Add ConstructorWithoutParamsCheck [issue 412]

zerg-junior:
This patch adds the ConstructorWithoutParamsCheck

GitHub:
 [sevntu.checkstyle issue 412] (https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/412)

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/ConstructorWithoutParamsCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/ConstructorWithoutParamsCheck.java
@@ -1,0 +1,180 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2016 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.sevntu.checkstyle.checks.design;
+
+import java.util.regex.Pattern;
+
+import com.puppycrawl.tools.checkstyle.api.Check;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.CommonUtils;
+
+/**
+ * This check prohibits usage of parameterless constructors,
+ * including the default ones.
+ *
+ * <p>
+ * <b>Rationale:</b> constructors of certain classes must always take arguments
+ * to properly instantiate objects. Exception classes are the primary example:
+ * their objects must contain enough info to find out why an exception occurred
+ * (see "Effective Java", item 63). Constructing an exception without a cause exception
+ * or an exception message leaves out such info and thus should be prohibited.
+ * </p>
+ *
+ * <p>
+ *     This check prohibits classes which simple names match the RegExp
+ *     defined in 'classNameFormat' property.
+ * </p>
+ *
+ * <p>
+ *     <b>Default configuration:</b>
+ *     <pre>
+ *         &ltmodule name="ConstructorWithoutParamsCheck"&gt
+ *           &ltproperty name="classNameFormat" value=".*Exception"/&gt
+ *           &ltproperty name="ignoredClassNameFormat" value="UnsupportedOperationException"/&gt
+ *         &lt/module&gt
+ *     </pre>
+ * </p>
+ *
+ * <p>
+ * <b>Examples:</b>
+ *
+ * <pre>
+ * // Assume a RegExp in classNameFormat catches example class names
+ * // the check can prohibit default constructors of built-in classes
+ * RuntimeException ex = new RuntimeException(); // violation expected
+ *
+ * // the check ignores classes which names match ignoredClassNameFormat
+ * // the default config ignores UnsupportedOperationException
+ * UnsupportedOperationException ex2 = new UnsupportedOperationException(); // no violation expected
+ *
+ * // the check allows constructors with empty arguments
+ * RuntimeException ex = new RuntimeException(""); // no violation expected
+ *
+ * // the check can prohibit default constructors of user-defined classes
+ * public class Clazz1 {
+ * }
+ *
+ * Clazz1 o1 = new Clazz1(); // violation expected
+ *
+ * // the check can prohibit  user-defined parameterless constructors
+ * public class Clazz2 {
+ *
+ *   Clazz2() {
+ *     foobar();
+ *   }
+ *
+ * }
+ *
+ * Clazz2 o2 = new Clazz2(); // violation expected
+ *
+ * </pre>
+ *
+ * </p>
+ * For more examples , see InputConstructorWithoutParamsCheck.
+ * For discussion, see the sevntu-checkstyle
+ * <a href=https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/412> issue 412</a>.
+ *
+ * @author <a href="mailto:Sergey Dudoladov@gmail.com">Sergey Dudoladov</a>
+ */
+public class ConstructorWithoutParamsCheck extends Check {
+
+    /**
+     * This key points to the warning message in the  "messages.properties" file.
+     */
+    public static final String MSG_KEY = "constructor.without.params";
+
+    /**
+     * The format string of the regexp for a check to apply to.
+     */
+    private String classNameFormat = ".*Exception";
+
+    /**
+     * The format string of the regexp of class names to ignore.
+     */
+    private String ignoredClassNameFormat = "UnsupportedOperationException";
+
+    /**
+     * The regexp to match against.
+     */
+    private Pattern regexp = CommonUtils.createPattern(classNameFormat);
+
+    /**
+     * The regexp to select class names to ignore.
+     */
+    private Pattern ignoredRegexp = CommonUtils.createPattern(ignoredClassNameFormat);
+
+    /**
+     * Sets the classNameFormat based on the XML configuration value.
+     *
+     * @param classNameFormat the regexp pattern
+     */
+    public void setClassNameFormat(String classNameFormat) {
+        this.classNameFormat = classNameFormat;
+        regexp = CommonUtils.createPattern(classNameFormat);
+    }
+
+    /**
+     * Sets the ignoredClassNameFormat based on the XML configuration value.
+     *
+     * @param ignoredClassNameFormat the regexp pattern
+     */
+    public void setIgnoredClassNameFormat(String ignoredClassNameFormat) {
+        this.ignoredClassNameFormat = ignoredClassNameFormat;
+        ignoredRegexp = CommonUtils.createPattern(this.ignoredClassNameFormat);
+    }
+
+    @Override
+    public int[] getDefaultTokens() {
+        return new int[] {TokenTypes.LITERAL_NEW};
+    }
+
+    @Override
+    public int[] getRequiredTokens() {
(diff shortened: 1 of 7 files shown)


zerg-junior:
The build fails because of the 94.4 % line coverage. Can I simply add this check to Cobertura exceptions ?

romani:
No, all new Check has to be 100% covered to prove that they are tested and there are no dead code

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/ConstructorWithoutParamsCheck.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public int[] getRequiredTokens() {
+        return getDefaultTokens();
+    }
+
+    @Override
+    public int[] getAcceptableTokens() {
+        return getDefaultTokens();
```
zerg-junior: @romani Cobertura reports this method call (line 120) to be the only one not covered by tests. Which code is supposed to call it ?  The user of this check is not supposed to set any other tokens besides LITERAL_NEW.

romani: 'getAcceptableTokens' should be called by checkstyle's Checker or Treewalker, most likely mocks of UT do not call them. Just call this method from UT to make coverage be happy.

zerg-junior:
@romani Updated to 100% coverage. This time the build failed with 
>  Report submission to Coveralls API failed with HTTP status 500: Internal Server Error (Build processing error.

Not sure if the error is related to this check.

romani:
looks like network problem, relauched.

zerg-junior:
@romani Can you please relaunch the build one more time ? It might have been some transient issue yesterday.

romani:
problem with "coverall" service, I try to relaunch build periodically

zerg-junior:
>  I try to relaunch build periodically
In the meantime, could you please review the code, so that I can fix issues ?

romani:
1) 
please make a reference to issue in commit message by as #XXXXX . Githib will parse it to a link to issue.

2)
>
```
<p>
+ *   This check prohibits usage of parameterless constructors,
```

please remove `<p>` for first line only, javadoc tread first line of comment in special way.

3)
In javadoc please specify that match by name is done by class SimpleName not by Canonical names.

4)
> `< module name`

please remove spaces between `<` and name of tag. In whole javadoc.

5)
> // the check can prohibit

we try to avoid word "prohibit" in documentation. you do not know a level of violation that will be in user config for your Check. we use "violation" "violated" terms. Please fix all usages.

6) 
>  private Pattern regexp = Pattern.compile(classNameFormat);

please reuse `CommonUtils.createPattern`

7)
> private void updateRegexp() {

too small to exists. please inline.

8)
> public void setClassNameFormat(String classNameFormat) {

please move up to be right after all fields.

9)
> numOfParameterExpr

no short names when you have a space. Rename to : parametersCount , numberOfParameters .

zerg-junior:
> we use "violation" "violated" terms. Please fix all usages.

Could you please point me to an example of this usage? The phrase "this check violates usage of parameterless constructors " does not sound right to me. Confusingly, other checks also use words like "forbid" and "warn".

zerg-junior:
>> < module name

> please remove spaces between < and name of tag. In whole javadoc.

Without the space,  this JavaDoc fails the build with the checkstyle error ` (javadoc) JavadocStyle: Extra HTML tag found: </module>`, although the opening  `<module>` tag is present.

romani:
> Could you please point me to an example of this usage?

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/coding/ModifiedControlVariableCheck.java#L41
https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.java#L71

just search .... there are a lot of examples.

>  Confusingly, other checks also use words like "forbid" and "warn".

yes, unfortunately, my bad to allow that, I missed this so many times.
"forbid" is too aggressive, and looks weird when severity is "info" and nobody care and pay attention
"warn" - is severity level. Check do not control severity. Check just do "log(......)"  - report violation. Severity is user configuration for each Check individually.

> Without the space, this JavaDoc fails the build with the checkstyle error

yes, such Check is buggy, we already have issues on it. Can you look around and to see how other javadocs do workaround of this problem?

zerg-junior:
@romani Updated

romani:
perfects code = simple  and clear.

1) last minor: 
> `* @author`

please make it single line.

2)
lets start testing :) on real code . Please use https://github.com/checkstyle/contribution/tree/master/checkstyle-tester tool. Attention to https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/my_check.xml#L15 (example of how sevntu Check could be used)
Generate reports against all projects. I doubt that you will find a lot of violations. Please share report at github.io , see "DEPLOY: " in README.md. 
Please do not hesitate to ask me any help, this tool is not ideal.
Please review all caught violations yourself first before sharing with me.

zerg-junior:
@romani  Please bump up the sevntu.checkstyle version at the [tester pom](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/pom.xml#L16) to 1.18.0
The tester currently cannot find `.class` file for some checks, including the one in the [example conf](https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/my_check.xml#L15)

> I doubt that you will find a lot of violations. 

I've found 745 warnings about exceptions with the empty parameter list. I'll export them in a couple of days.

romani:
> Please bump up the sevntu.checkstyle version 

done.

zerg-junior:
@romani [List of violations](https://zerg-junior.github.io/)

402 out of 745 violations are caused by `UnsupportedOperationException`. I personally think this exception  should provide description of what prevents implementation, e.g. say that a method that throws the exception is optional and thus not implemented.   However, such  widespread usage of this Exception without any parameters makes me question this judgement.

82 come from `NoSuchElementException`. Here I feel the cause should be included, preferably with some boundary conditions.

PS Please update the "Deploy" section of the tester docs to mention that `target/site/checkstyle.html` should be manually renamed to `index.html`, otherwise GitHub Pages won't display the site. The current `index.html` file just links to `checkstyle.html`.

romani:
> PS Please update the "Deploy" section 

Done, thanks a lot for pointing to this.

zerg-junior:
@romani What should happen to the violation report ?

romani:
Thanks a lot, all done correctly, just give me some time to review it.

romani:
> List of violations

too much violations, it is usually a sigh that Check is too strict and in real live is not practical.

1)
Please provide report against Spring, openJDK7 and hibernate.
If amount of violations is the same - it is sign that Check need to have "exclude" option to let user decide what names should be skipped (example: "Test*Exception|NullPointerException|UnsupoportedoperationException").


2) please make a PR to guava project, with update for 10 cases and lets see their reaction on this change. We already did this, sometimes they merge, sometime they describe a reason why it should not be done.

zerg-junior:
Please upgrade the [BaseCheckTestSupport](https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/test/java/com/github/sevntu/checkstyle/BaseCheckTestSupport.java) to use checkstyle 6.16
I run into [this issue](https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/412#issuecomment-182795792) again and had to roll back to 6.11.2.

Even worse, after the roll back I started to observe this issue:
` src/main/java/com/github/sevntu/checkstyle/checks/design/StaticMethodCandidateCheck.java:[192,23] (coding) FinalLocalVariable: Variable 'frame' should be declared final.`
When I try to fix it, the compiler complains
`Compilation failure: Compilation failure: sevntu.checkstyle/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/StaticMethodCandidateCheck.java:[218,17] cannot assign a value to final variable frame`

> provide report against Spring, openJDK7 and hibernate.

I caught a bug in the check while preparing the reports for openJDK7 and hibernate. The bug does not manifest while checking the Guava codebase. In the future, sevntu checks should probably be tested against  several different projects by default.

romani:
> Please upgrade the BaseCheckTestSupport to use checkstyle 6.16

done.

> I caught a bug in the check while preparing the reports for openJDK7 and hibernate.

:) , a lot of people blaming me to be over protective in testing, now you see a reason why I am so demanding.

>In the future, sevntu checks should probably be tested against several different projects by default. 

yes, but we do not have that much contributors to make it happen. But we slowly going into this direction: https://github.com/checkstyle/checkstyle/wiki/Checkstyle-GSoC-2016-Project-Ideas#project-name-regression-testing-tool-and-html-report-generator-for-pull-request , https://github.com/checkstyle/checkstyle/wiki/Checkstyle-diff-report-genertion-service , https://github.com/checkstyle/contribution/issues . New contributors are welcome to help us do auto-testing.

zerg-junior:
@romani [Violations for Guava, openJDK, hibernate and Spring](https://zerg-junior.github.io/)

I generated the report with the code that for some reason fails to build on GitHub, so please point me explicitly to the issue that causes the failure; the build trace seems to be Ok and then suddenly fails.

> I caught a bug in the check

The [bug fix](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/428/files#diff-9996e5482dfb8038c6945544b3c89d64R139) and relevant [code samples](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/428/files#diff-40a17bd0e9a6bad116cd7523260e0684R42).

>> Can you please confirm if other sevntu.checkstyle installations have this problem with checkstyle 6.11.2 ?

> please rephrase a question , I do not understand it.

Never mind, this was misconfiguration on my side.

romani:
>  so please point me explicitly to the issue that causes the failure; the build trace seems to be Ok and then suddenly fails.

I do not understand, but CI is green, all is fine.


------------------

wget https://zerg-junior.github.io/ -O report.html && html2text -o report.txt report.html
grep "ConstructorWithoutParams" report.txt | sed -E "s/.*Params '//" | sed -E "s/' .*//" | sed -E "s/.*Calls to constructors of '//"  | sort | uniq -c

romani:
@zerg-junior , look like we need to make an property in Check to skip some exceptions (regexp type). By default skip ONLY "UnsupportedOperationException". By your report is clear that developers do not like comment it as what unsupported operation is mentioned in exception stacktrace.

Additionally this property will give user freedom to define himself how what exceptions could skip description. The most fanatical team could make this option as non-matching regexp and force all Exceptions to have description.

zerg-junior:
@romani [Updated code](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/428/files#diff-9996e5482dfb8038c6945544b3c89d64R168), [updated report](https://zerg-junior.github.io)

BTW, there seem to be a bug in the tester: once source code of a project is downloaded, this project will always checked no matter what `projects-to-test-on.properties` says.

romani:
>BTW, there seems to be a bug in the tester: once source code of a project is downloaded, this project will always be checked no matter what 

It is expected for now, this kind of desire to skip download when you/we test of the same projects. But yes ,I even myself run into this problem leftovers in src folder of checkstyle-tester. if you can prove that this is more inconvenience than benefit - I am ok to change behaviour. For now I do not know what is better.

romani:
report is good, thanks a lot.

last minors before merge:

1)
> /* The "new" keyword eith

please use single line comments inside a method

2)
>```
 +       /*
+        * RuntimeException ex = new RuntimeException();
+        */
```

if this is for testing of comments, please put a comment. It looks like a dead code.
Only certain set of Checks work on a tree with comments, your Check is unaware by design.

zerg-junior:
@romani Updated.

romani:
thanks a lot.

sorry that I did not mention this before but 

please update all other configuration files
see examples at https://github.com/sevntu-checkstyle/sevntu.checkstyle/commit/c8b9f766238add3899cba11b42d05ad039e773b7 (eclipse-cs, sonar, .... )

please update https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/gh-pages/sevntu-checkstyle-default-configuration.xml to have your Check with all properties, it is used as example.

zerg-junior:
@romani Updated.

romani:
> default-configuration.xml to have your Check with all properties, it is used as example.

please do separate PR for this file, it is located in gh-pages branch.

romani:
@zerg-junior , example of update for gh-pages file https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/443

zerg-junior:
@romani  [Pull request for gh-pages](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/444/files#diff-066db4e6a5ba6711ee490af9080d4ff6R203)

romani:
merged as FF.

thanks a lot for your contribution !!!

zerg-junior:
@romani  Many thanks for rebasing the gh-pages request and merging :)

Issue #425, opened 2016-01-18
NumericLiteralNeedsUnderscore's new option to skeep variables by name


romani:
all details : https://travis-ci.org/checkstyle/checkstyle/jobs/103024859

code coverage should be 100%, so line https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/pom.xml#L134 should be removed.

romani:
@cypai , please address this ASAP.

please generate reports (by chekcstyle-tester project) over other projects to make sure there are no other code that need to be ignored.

cypai:
Will be done this week

cypai:
@romani 
I think exactly which variables to be ignored merits some discussion. Like, should only variable declarations be ignored? For example:

```java
public static final long serialVersionUID = 1234567890L;  // This is ignored

public void bar() {
  long foo = 1234567890L;  // This is ignored
  foo -= 1234567;  // What about this?
  foo = 1234567;  // What about this?
}
```

romani:
We need to ignore variable or field that could have generated value. For now we know only about fields that could have generated fields that people usually do not care but that used in massive amount of code.

For now, lets make ignore option only for field names, serialVetsionID is default value.

After update please run Check against some opensource code to make sure it works and nothing new need to be created. You can use contribution/checkstyle-tester project to generate html report.

cypai:
From discussion:
Ignore configuration should be called ignoreFieldNamePattern. It should be done as a Regexp.
These should only care about class fields; if a number is used in a function then it is significant and should be checked.
Reference the main checkstyle project for examples of getting fields.

Afterwards, run it on Spring, Guava, and Orekit to see what comes up.

romani:
default value of ignoreFieldNamePattern is "serialVersionUID" .

Try to run your Check at your private projects.

romani:
@cypai , please finish this issue.

romani:
@cypai , please send PR checkstyle repo to start usage of this Check.
I will do release of sevntu tomorrow.

FYI: All senvtu Checks have to be used in main repo - checkstyle. config - https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle_sevntu_checks.xml

romani:
@cypai ,

New option is introduces  but eclipse-cs/sonar/default-config are not updated:

[...]

Issue #108, opened 2013-04-06
New Check: ForbidParameterInAnnotation


romani:
Use case:
usage of "expected" parameter in annotation Test (http://junit.sourceforge.net/javadoc/org/junit/Test.html ) is not convenient for UnitTesting as it prints exceptions in logs so it will be distracting during UTs failures examination in Jenkins or during maven build. It is hard to distinguish expected Exception and unexpected Exception that appear in test but due to lack of appropriate check does not fails. 

Rule of thumb: no exceptions means all is ok, exception appear - it is a problem.

So we need to create Check that will allow us to detect parameter usage in in required Annotation.
Options:
"annotationName" is simple String
"parameterName" is simple String
"parameterValue"  is regexp String, default value is ".*"

romani:
Sergey does not have time to finish , but 90% of work is done:
https://github.com/drozzds/sevntu.checkstyle/commit/cca522efd875dcc913e6169d3c3ff963a09e1cf3

non finished discussion:
https://groups.google.com/d/msg/sevntu-checkstyle/5MFHVlFTmz8/YBuBK2IonZsJ

romani:
fix is merged.

romani:
@rnveach , issue to use new Check - https://github.com/checkstyle/checkstyle/issues/3567

romani:
@rnveach , we missed update at https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checkstyle-sonar-plugin/src/main/resources/com/github/sevntu/checkstyle/sonar/checkstyle-extensions.xml

rnveach:
Update in PR #487

PR #431, opened 2016-03-05
Issue #425: New option to skip fields by regexp

cypai:
#425

### DIFF
--- a/sevntu-checks/pom.xml
+++ b/sevntu-checks/pom.xml
@@ -23,6 +23,18 @@
       <version>4.10</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-api-mockito</artifactId>
+      <version>1.6.4</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-module-junit4</artifactId>
+      <version>1.6.4</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <distributionManagement>
     <repository>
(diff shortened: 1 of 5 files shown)


romani:
@cypai ,

1)
> logWarningIfAstDoesNotPass

if try to name method as "xxxxxxIfxxxxxxxxxx" that is big sign that smth is wrong in design.
Please inline this method to visitToken

2)
> 
```
 +        final boolean tokenIsField = isField(ast);
+        if (tokenIsField) {
+            if (!ignoreFieldNamePattern.matcher(getFieldName(ast)).find()) {
```

please move this code to `passCheck` method.

3)
>
` <regex><pattern>.*.checks.coding.NumericLiteralNeedsUnderscoreCheck</pattern><branchRate>95</branchRate><lineRate>93</lineRate></regex>`

please remove this line from pom.xml and make coverage 100%.

cypai:
The only remaining non-covered lines are the default lines in the switch statements. Any suggestions?

romani:
unfortunately, the only way to make it is to convert SWITCH to IF_ELSE and use reflection to get access to private methods.

Example:
code: https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.java#L362
UT: https://github.com/checkstyle/checkstyle/blob/master/src/test/java/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheckTest.java#L420

but benefit of 100% coverage will be much better than such verbose and non-optimal structure and hacky UT.

cypai:
What's the benefit for these last lines of coverage? I don't see any real benefits, plus this if-else structure is far less readable. It doesn't seem worth writing such hacky UTs that depend on internal implementation.

romani:
> What's the benefit for these last lines of coverage?

the most valuable reason - absolutely no dead code. With coverage close to 100% there is possibility to introduce new code and miss coverage of it. I do not want to keep an eye on each code to be covered. That should be enforced by a coverage tool during build. Whole code in main checkstyle library is 100% covered. Benefit is enormous - I do not review coverage reports at all - they are 100% always. Any new code should be covered by UT.
 
Unfortunately coverage tools are not ideal and some design approaches are not so easy to mock Life is not ideal. We do hack (reflection) in UT area, so main code is not affected. Unfortunately your case is most problematic and to reach 100% we need to change main code. But slightly verbose code in comparison to threat to lose coverage is minor. 

After you got 100% coverage, nobody will be allowed to lower it. Code will be fully testable forever.

romani:
@cypai , line from  sevntu-checks/pom.xml should be removed completely.

cypai:
@romani Done.

romani:
Usage of `@Test(expected=)` is highly discouraged it make test non-strict and just expect some Exception to happen .... please send separate PR with fix for this

romani:
hmmm, https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/108 still not done ....... that is why it is not reported.

Issue #446, opened 2016-03-28
ConstructorWithoutParamsCheck: update the default config


zerg-junior:
Currently the check with the default config matches the pattern ".*Exception" anywhere in a class name. This leads to false positives such as in "MockExceptionHandler".

zerg-junior:
[Pull request](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/447)

rnveach:
Closing as fix was merged in referenced commit

PR #447, opened 2016-03-28
 Issue #446: Udpdate the default config in ConstructorWithoutParamsCheck

zerg-junior:


romani:
thanks a lot ! (sorry for delay with merge).

I think we do not need hotfix release.

PR #452, opened 2016-04-28
Add 'mvn javadoc:javadoc' Travis step for all projects left

daniilyar:


romani:
it is possible to do smth like "mvn javadoc:javadoc verify" ?

daniilyar:
> it is possible to do smth like "mvn javadoc:javadoc verify" ?
yes, it is. I will update travis config appropriately soon

PR #456, opened 2016-05-21
minor: updated eclipse formatter to checkstyle current standards

rnveach:
Done on Eclipse Luna SR 2 on Windows.
I imported the original, went through a few files and played around with the settings to get them close to how CS is currently formatted.
Not sure if we want to validate this somehow...

Eclipse doesn't save the file in the same order, so I created a small program to re-sort it based on the original XML sorting.


rnveach:
@romani ping

romani:
Merged. Thanks a lot for pinning me.

PR #465, opened 2016-07-29
Check for trailing commas on enums

kariem:
This checks that the last enum constant definition in a multi-line enum ends in a comma:

    enum E1 {
        ONE,
        TWO,
        THREE,
    }

Addresses #464

PR #466, opened 2016-08-01
Replace `Assert` class from internal Sun package with junit `Assert`

kariem:
This prevented me from running the tests in the IDE. Seems someone has used the wrong `Assert` class.


romani:
good catch !

kariem:
Thank you!

The build process contains a lot of very strict rules, but then something like this just slips through. For the other PR I've opened earlier (#465), I had to make the code a little less readable to get to the 100% line and branch coverage. For new code, I think that is good, however, we should also look at existing code to improve the situation.

romani:
We reached 100% for existing code in main project, for this experimental project we do not have time. But everyone are welcome to improve.

PR #477, opened 2016-11-16
.classpath with minor additions, as per M2E

vorburger:
The latest m2e (Maven support for Eclipse) creates the .classpath file
like this.  It's a PITA that it always shows up as "dirty" in Git, and
it would thus be best to merge this to avoid that.

An alternative could be to not keep Eclipse metadata in Git at all
anymore.

Change-Id: I91d4417183fa6d5a9db2b4fa4049b1e0fbfef04f
Signed-off-by: Michael Vorburger <vorburger@redhat.com>

### DIFF
--- a/sevntu-checks/.classpath
+++ b/sevntu-checks/.classpath
@@ -34,6 +34,10 @@
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
 	</classpathentry>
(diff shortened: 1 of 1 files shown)


vorburger:
@romani willing to merge this minor contribution?

romani:
> An alternative could be to not keep Eclipse metadata in Git at all anymore.

it might be good, we succeeded to do this in main project, but not in sevntu.
If you know how to make it - contribute.

PR #479, opened 2016-11-16
dos2unix to convert 4 files 'with CRLF line terminators' (Windows instead of Linux EOL)

vorburger:


vorburger:
@romani this will make the bad diff in https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/480/files#diff-c0b4ea0807164ef9a0ac5e219d0d489b disappear. I'll just send another PR for checkstyle_checks.xml to enforce this, if you're interested in that.

Issue #461, opened 2016-06-30
New Check: Logger might be declared only once


vorburger:
As per https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/457:

_Describe its behaviour with examples of code, with violation:_

    org.slf4j.Logger logger1;
    org.slf4j.Logger logger2;

_and without:_

    org.slf4j.Logger logger;

_Examples on how configuration will look like:_

This new Check initially will have no configuration.

romani:
ok, please update description and show how you plan configuration will looks like. And put explanation on your strategy to distinguish required type among other types.

vorburger:
@romani updated description - OK like this?

romani:
Check that is not configurable ... not good. 
LoggerDeclarationsCountCheck is to general name in this case, not acceptable and too misleading.


why not to use special property to let user define fully qualified class name ?
I have never see a code that used fully qualified types for loggers ...what if logger field is `Logger logger2;` (with import before)

vorburger:
> Check that is not configurable ... not good.
> why not to use special property to let user define fully qualified class name ?

Sure, if you like; I've adapted the example with a <property> for this on top.

> LoggerDeclarationsCountCheck is to general name in this case, not acceptable and too misleading.

Sure, so what would you name it? Now that it's configurable, it actually has nothing to do with Logging anymore... so.. maybe just FieldDeclarationsCountCheck, do you like that? If not, then proposing a concrete alternative instead of only "not acceptable" would be great.

> I have never see a code that used fully qualified types for loggers ...what if logger field is Logger logger2; (with import before)

Yeah sure, obviously that works as well, see how they (someone else, not me) have done it here https://github.com/sevntu-checkstyle/sevntu.checkstyle/compare/master...vorburger:issue457a_LoggerDeclarationsCountCheck#diff-c80463aaf2d1a96b49371088ce995266R49 ... I've adapted the example for this on top to show both.

romani:
> com.github.sevntu.checkstyle.checks.logging.FieldDeclarationsCountCheck

name is better.
please fix, package name

> // both FQN and import'ed type name works

please update code sample to be real, ready to copy-paste and validate by checkstyle, by trailing comment please mark lines that will have violation.

vorburger:
> please fix, package name

You mean in probably .coding. instead of my propose new .logging. ? Sure, done.

> please update code sample

[...]

Issue #457, opened 2016-05-25
Contributing (existing) slf4j Logger Check


vorburger:
Hello, as per https://lists.opendaylight.org/pipermail/yangtools-dev/2016-May/001391.html, we would be interested in contributing https://git.opendaylight.org/gerrit/gitweb?p=yangtools.git;a=tree;f=common/checkstyle-logging/src/main/java/org/opendaylight/yangtools/checkstyle;hb=HEAD.

This is a collection of individual new Checkstyle Checks related to slf4j logging, helping with enforcement of anti patterns such as:

* Logger must be declared as private static final
* Logger name should be LOG
* LoggerFactory.getLogger Class argument is incorrect
* Logger might be declared only once
* Logger must be slf4j
* Log message placeholders count is incorrect
* Log message contains string concatenation

Q1: Would you welcome this contribution?

Q2: How do we go about this, concretely? Just raise a Pull Request? Do you have any particular "process" ? Any special "quality requirements"  ;-) beyond [test coverage which we already have](https://git.opendaylight.org/gerrit/gitweb?p=yangtools.git;a=blob;f=common/checkstyle-logging/src/test/java/org/opendaylight/yangtools/checkstyle/CheckstyleTest.java;h=977b61a9c9ce5fe46f3b642338e893cb01ec4e61;hb=HEAD)

Q3: One of our motivations to contribute this is to avoid [having to write/build/distribute a custom Eclipse plugin](http://eclipse-cs.sourceforge.net/#!/custom-checks) just for these few Checks. Once this is contributed, would we have to do anything special to get these checks into the sevntu.checkstyle Eclipse plugin, or would you be able to help with that? 

Thanks.

romani:
> Q1: Would you welcome this contribution?

Yes, but only if idea is not so specific and are not false-positive prone.
Lets go with case by case. 

>  How do we go about this, concretely?

As usually: open issue with all details, let me review it and confirm. Raise a PR, pass code review, provide testing. Done :) . 

> Any special "quality requirements" 

 code review, test reports the same as in checkstyle project. One day all such Checks (after testing period) will be ported to checkstyle .

> would you be able to help with that? 

sevntu-checkstyle is official sandbox project to let people implement Checks and test them on real code for some time before introducing it to main project and and get storm of issue.
[...]

PR #480, opened 2016-11-16
Issue #461: FieldDeclarationsCountCheck (NEW)

vorburger:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/461


vorburger:
@romani I've tried, but failed, to understand what it's trying to tell me with the following violation in a new class I introduced - the exact same import appears in many other *Check classes - what on earth is it complaining about that I'm importing wrong here - can you help?

`[ERROR] src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheck.java:[22,1] (imports) ImportControl: Disallowed import - com.github.sevntu.checkstyle.Utils`

rnveach:
@vorburger That is a check that prevents certain packages from using what he has defined as illegal imports.
The violation is referring to this line: https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/480/commits/0f7539fd7ad1cab53542aebb57efe9bbb43cf277#diff-1113b12ada701bc205c9c2de790bda2aR22
Currently only `com.github.sevntu.checkstyle.checks.api` package is allowed to use that import. I am unfamiliar with the reasoning for this. It could be an oversight for all I know.

It looks like you added new methods to `Utils`. Do they need to be there instead of inside FieldDeclarationsCountCheck?
Also, you'll need to rebase your class to remove the conflicts.

vorburger:
> @vorburger That is a check that prevents certain packages from using what he has defined as illegal imports. 

http://checkstyle.sourceforge.net/config_imports.html#ImportControl

> It means you shouldn't be importing that class in FieldDeclarationsCountCheck.
> The violation is referring to this line: 0f7539f#diff-1113b12ada701bc205c9c2de790bda2aR22

Sure, I got that, but as I asked above, "the exact same import appears in many other *Check classes", see e.g. https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EmptyPublicCtorInClassCheck.java#L7, so why is that allowed there but in my new class it's whining about it?

> Currently only com.github.sevntu.checkstyle.checks.api package is allowed to use that import. I am unfamiliar with the reasoning for this. It could be an oversight for all I know.

There is no com.github.sevntu.checkstyle.checks.api package; thus proposing https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/484

> It looks like you added new methods to Utils. 

Yeah. I'm therefore proposing to solve this confounding "problem" by adding https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/480/files#diff-42857ffc420c808e64abbf2a0e0440b4R13

> Do they need to be there instead of inside FieldDeclarationsCountCheck?

Yes, because I'll be re-using those util methods in other changes from https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/457 I'm about to propose, as soon as we can get this one agreed upon and merged.

An alternative to sharing those as static methods in that Utils class would be to introduce a new AbstractFieldCheck kinda class and put those there (as non-static). But given that there already is a Utils class, IMHO it's better to use that instead of class hierarchy for this purpose.

> Also, you'll need to rebase your class to remove the conflicts.

Sure, done.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheckNOK.java
```diff
@@ -0,0 +1,12 @@
+package com.github.sevntu.checkstyle.checks.coding;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FieldDeclarationsCountCheckNOK {
```
romani: please add examples of usage in classes that have subclasses with loggers.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheckNOK.java
```diff
@@ -0,0 +1,12 @@
...
+
+    // both FQN and import'ed type names work:
+    Logger logger1;
```
romani: please make a test case with "Logger" from different(not slf4j) package in imports.
there should be no violation.
Check have to process imports to know what short name is actually used, we have examples in other Checks for this.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/Utils.java
```diff
@@ -69,4 +72,48 @@ public static DetailAST getNextSubTreeNode(DetailAST currentNode, DetailAST subT
         }
         return toVisitAst;
     }
+
+    /**
+     * Obtain full type name of first token in node.
+     * @param ast an AST node
+     * @return FQN of type
+     */
+    public static String getTypeName(final DetailAST ast) {
```
vorburger: OK, will rename it to getTypeNameOfFirstToken(), does that work better for you?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/Utils.java
```diff
@@ -69,4 +72,48 @@ public static DetailAST getNextSubTreeNode(DetailAST currentNode, DetailAST subT
...
+        final FullIdent ident = CheckUtils.createFullType(ast.findFirstToken(TokenTypes.TYPE));
+        return ident.getText();
+    }
+
+    /**
+     * Check node for matching class, taken both FQN and short name into account.
```
vorburger: OK, it's fully qualified name, I'll update the JavaDoc

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheck.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Checks that classes have at most 1 field of the given className.
+ *
+ * <p>This can be useful for example to ensure that only one Logger is used.
+ *
+ * @author Milos Fabian, Pantheon Technologies - original author (in OpenDaylight)
+ * @author Michael Vorburger.ch - refactored and made more generalized for contribution to Sevntu
+ */
+public class FieldDeclarationsCountCheck extends AbstractCheck {
```
vorburger: OK, renamed to FieldSingleDeclarationCheck in next iteration.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheck.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Class name check for.
+     */
+    private String className;
+
+    /**
+     * Field to hold previously seen Logger.
+     */
+    private String prevSeenClassName = "";
```
vorburger: OK, thanks a lot for pointing this out; done in next iteration.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheck.java
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Set Class of which there should only be max. 1 field per class.
+     * @param className the fully qualified name (FQN) of the class
+     */
+    public void setClassName(String className) {
+        this.className = className;
+    }
+
+    @Override
+    public int[] getDefaultTokens() {
```
vorburger: OK, if you say so.. I've just implemented getRequiredTokens() and getAcceptableTokens() as return getDefaultTokens(). There is no setTokens(). If that is NOK, please update that documentation to clarify, or point me to an example in another existing *Check to follow?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheck.java
```diff
@@ -0,0 +1,91 @@
...
+
+package com.github.sevntu.checkstyle.checks.coding;
+
+import com.github.sevntu.checkstyle.Utils;
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+/**
+ * Checks that classes have at most 1 field of the given className.
```
vorburger: OK, done.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheckNOK.java
vorburger: NOK, because my understanding is that Checkstyle cannot handle this, because its based on analysing the AST of a single file; AFAIK I couldn't check parent class fields, right? So.. no point, I would have thought.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheckNOK.java
vorburger: NOK, I do not know how to do this.. can you point me to an existing Check with a corresponding Test that does this, so I can copy and learn from it? This seems a lower priority to me - perhaps something that could be dealt with in a follow-up, after this is accepted and merged?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Override
+    public void beginTree(DetailAST rootAST) {
+        prevSeenClassName = null;
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        if (isVariable(ast) && Utils.matchesFullyQualifiedName(ast, className)) {
```
rnveach: You are not looking at import statements. User can import class `Logger` and it not be from `org.slf4j`, but you would still mark it as violation.
Why do I need to define full path in check (`org.slf4j.Logger`), as your examples show, if check won't obey it? You could just give `Logger` only and it will be same result.
@romani your thoughts?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
```diff
@@ -0,0 +1,120 @@
...
+            final String nowVisitingClassName = Utils.getClassName(ast);
+            if (this.prevSeenClassName != null
+                    && this.prevSeenClassName.equals(nowVisitingClassName)) {
```
rnveach: @vorburger This line puzzles me.
I am not sure when `prevSeenClassName` will not equal `nowVisitingClassName`. User gave specific class, so it should always be said class. I assume we are passing line/branch coverage here.
Can you point me to instance where `equals` evaluates to false?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Class name check for.
+     */
+    private String className;
+
+    /**
+     * Field to hold previously seen Logger.
+     */
+    private String prevSeenClassName = "";
```
vorburger: OK, done, using empty string instead of null consistently everywhere now.

vorburger:
@rnveach re. "Please look into full file difference in checkstyle-extensions.xml. I think we should only see your additions.", this will be fixed by https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/490; unfortunately once again only one-off. In an ideal world, we really should do https://github.com/checkstyle/checkstyle/issues/893 so we can do https://github.com/checkstyle/checkstyle/pull/3557

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
vorburger: @rnveach you're right, I think (from what I understand of CS). But this almost more of a limitation of CS.. I had a quick look at EitherLogOrThrowCheck, and see how that Check has code to deal with this (hasLoggerClassInImports), but... that almost looks a like a one off hack to me. If you would like Checks to be able to be written for both full and local type names, then shouldn't there be better shared infrastructure available for writing such Checks? I would argue that mixed types of Logger is extremely atypical, and so while I agree this isn't completely perfect strictly speaking, it seems like a very minor problem in real world, and we should ignore this for now, and move forward. @romani your thoughts?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
vorburger: OK, I've inherited this code from someone else. This is a left over from when it didn't reset in beginTree and finishTree; so it kept track of which class its visiting. Meanwhile, I guess this could be written differently now - could you please have a look at the latest iteration - makes sense?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
```diff
@@ -0,0 +1,120 @@
...
+            if (this.prevSeenClassName != null
+                    && this.prevSeenClassName.equals(nowVisitingClassName)) {
+                log(ast.getLineNo(), MSG_KEY, className);
+            }
+            this.prevSeenClassName = nowVisitingClassName;
+        }
+    }
+
+    @Override
+    public void finishTree(DetailAST rootAST) {
```
vorburger: OK, removed.

sevntu-checks/src/main/resources/com/github/sevntu/checkstyle/checks/coding/messages.properties
```diff
@@ -33,6 +33,7 @@ either.log.or.throw=Either log or throw exception.
 explicit.init=Variable ''{0}'' explicitly initialized to ''{1}'' (default value for its type).
 fall.through=Fall through from previous branch of the switch statement.
 fall.through.last=Fall through from the last branch of the switch statement.
+field.count={0} might be declared only once.
```
vorburger: OK, I think that was meant to be an imperative "may" instead of "might", but "should" is better; done.

vorburger:
@romani could you please advise how to address the coverage problem shown below, which is due to the introduction of `getAcceptableTokens()` you've requested earlier above here.. I've no idea what that does, and don't see what one would have to do in the test to make that covered - can you point to a line in an existing test which leads to coverage of `getAcceptableTokens()` ?  This is required because you seem to (now, new; since you've fixed Cobertura) be enforcing 100% test coverage.

`[ERROR] com.github.sevntu.checkstyle.checks.coding.FieldSingleDeclarationCheck failed coverage check. Line coverage rate of 93.7% is below 100.0%`

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/Utils.java
```diff
@@ -74,4 +77,63 @@ public static DetailAST getNextSubTreeNode(DetailAST node, DetailAST subTreeRoot
...
+     * Obtain full type name of first token in node.
+     * @param ast an AST node
+     * @return fully qualified name (FQN) of type, or null if none was found
+     */
+    public static String getTypeNameOfFirstToken(final DetailAST ast) {
+        final DetailAST findFirstToken = ast.findFirstToken(TokenTypes.TYPE);
+        if (findFirstToken != null) {
+            final FullIdent ident = CheckUtils.createFullType(findFirstToken);
+            if (ident != null) {
+                return ident.getText();
```
rnveach: It is already an open issue in main project where sevntu get its configuration from.
Just no one has worked on it.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
rnveach: New iteration definitely makes more sense to me.

rnveach:
> how to address the coverage problem shown below, which is due to the introduction of getAcceptableTokens()
@vorburger 
`getAcceptableTokens` is [called in TreeWalker when it is first registering the check](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/TreeWalker.java#L214).
Every check accepts a property `tokens`. It is how we allow users to customize what some check validates pertaining to specific nodes.

Try adding something like `config.addAttribute("tokens", "CLASS_DEF");` to a new test.

vorburger:
@rnveach thank you, that seems to work; done.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        if (fullyQualifiedClassName == null) {
+            throw new IllegalStateException("Must set mandatory className property in"
+                    + getClass().getSimpleName());
+        }
+        if (Utils.matchesFullyQualifiedName(ast, fullyQualifiedClassName)) {
+            if (hasPreviouslySeenClass) {
```
romani: I do not see filtering for fields.
Your Check is named - FieldSingleDeclarationCheck , but you examine fields+variables.

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheckTest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Test
+    public void testFieldSingleDeclaration_OK() throws Exception {
+        String className = "org.slf4j.Logger";
+        config.addAttribute("className", className);
+        verify(config, getPath("FieldSingleDeclarationCheck_OK.java"), new String[] {});
+    }
+
+    @Test(expected = CheckstyleException.class)
```
romani: @rnveach , I thought we already made a Check that control that bad practice.
sevntu-checkstyle.github.io/sevntu.checkstyle/apidocs/com/github/sevntu/checkstyle/checks/annotation/ForbidAnnotationElementValueCheck.html
Do we miss to add it to config ?

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck_OK.java
```diff
@@ -0,0 +1,10 @@
+package com.github.sevntu.checkstyle.checks.coding;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FieldSingleDeclarationCheck_OK {
+
+    org.slf4j.Logger logger;
+
+}
```
romani: I do not see tests for cases with inner classes (in class , in method, ..... yes real life code is crazy on purpose).
If we care only about Top level class we should define this in javadoc or .... .

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/Utils.java
romani: ok, for now

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheck.java
romani: I agree that our documentation is not ideal .... . Unfortunately usage of them is different for each Check, in some Check all of them return different in size array of Tokens (but always interchangeable). 
Can you give us advices on how you would like too see this doc (in separate issue) ? For now I am blind as I spent too much time with that project.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheck.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public int[] getDefaultTokens() {
+        return new int[] {TokenTypes.VARIABLE_DEF };
+    }
+
+    @Override
+    public void visitToken(DetailAST ast) {
+        if (isAFieldVariable(ast) && Utils.isMatchingClass(ast, className)) {
```
romani: keep thinking like this on my each future requests :) , I know I over-demanding for quality.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheckNOK.java
romani: sorry, I meant inner classes. I did a comment one more time in my second iteration of review.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/FieldDeclarationsCountCheckNOK.java
romani: Examples:
```
~/java/github/sevntu-checkstyle/sevntu.checkstyle/sevntu-checks/src/main/java [master|✔] 
$ grep -R "IMPORT" * | head -n 5 
com/github/sevntu/checkstyle/checks/coding/ForbidInstantiationCheck.java:        return new int[] {TokenTypes.IMPORT, TokenTypes.LITERAL_NEW };
com/github/sevntu/checkstyle/checks/coding/ForbidInstantiationCheck.java:            case TokenTypes.IMPORT:
com/github/sevntu/checkstyle/checks/coding/ForbidInstantiationCheck.java:     *        literal node ("IMPORT" or "LITERAL_NEW" node types).
com/github/sevntu/checkstyle/checks/coding/ForbidInstantiationCheck.java:     *         "IMPORT" node or instanstiated class Name&Path for given
com/github/sevntu/checkstyle/checks/coding/ForbidCertainImportsCheck.java:                TokenTypes.IMPORT, TokenTypes.LITERAL_NEW, };
```
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ForbidInstantiationCheck.java

Yes, it could be postponed but in this case please mention this in javadoc as limitation that could be improved in future.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Violation message key.
+     */
+    public static final String MSG_KEY = "field.count";
+
+    /**
+     * Class name check for.
+     */
+    private String className;
```
romani: Ideally Check should be more general than default configuration.
If default configuration is not possible to make as reasonable, Check should do nothing and no crashes are expected.

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheckTest.java
rnveach: @romani See my comment above, sevntu doesn't run its own checks against itself. Only Checkstyle checks.
Plus, we don't run anything against Test in sevntu hence why there is no header in this file.

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheckTest.java
```diff
@@ -0,0 +1,59 @@
...
+public class FieldSingleDeclarationCheckTest extends BaseCheckTestSupport {
+
+    private final DefaultConfiguration config = createCheckConfig(FieldSingleDeclarationCheck.class);
+
+    @Test
+    public void testFieldSingleDeclaration_NOK() throws Exception {
+        String className = "org.slf4j.Logger";
+        config.addAttribute("className", className);
+        final String expected[] = { "10: " + getCheckMessage(MSG_KEY, className) };
+        verify(config, getPath("FieldSingleDeclarationCheck_NOK.java"), expected);
```
rnveach: Input files should start with `Input`.

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheckTest.java
```diff
@@ -0,0 +1,59 @@
...
+        verify(config, getPath("FieldSingleDeclarationCheck_OK.java"), new String[] {});
```
rnveach: One test per method please.

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheckTest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Test
+    public void testFieldSingleDeclaration_FQN_NOK() throws Exception {
+        String className = "org.slf4j.Logger";
+        config.addAttribute("className", className);
+        final String expected[] = { "7: " + getCheckMessage(MSG_KEY, className) };
+        verify(config, getPath("FieldSingleDeclarationCheck_FQN_NOK.java"), expected);
+        verify(config, getPath("FieldSingleDeclarationCheck_OK.java"), new String[] {});
```
rnveach: Same here, one test per method.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
romani: >  shared infrastructure available for writing such Checks?

Yes we lack a lot on infrastructure util methods and examples and .... our team is very very tiny.
All we have is what you see in repo, all re welcome to improve :) .

> it seems like a very minor problem in real world, and we should ignore this for now, and move forward.

we could, I am ok with multi-step development. All depends on your time.
But on each step logic should be clear for user. So if we agree on limitation, it should be in javadoc and user should clear see this, in other case there will be storm of requests/bugs from users.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/FieldSingleDeclarationCheck.java
```diff
@@ -0,0 +1,107 @@
...
+     * Field to remember if class was previously seen in current File.
+     */
+    private boolean hasPreviouslySeenClass;
+
+    /**
+     * Set Class of which there should only be max. 1 field per class.
+     * @param className the fully qualified name (FQN) of the class
+     */
+    public void setClassName(String className) {
+        this.fullyQualifiedClassName = className;
```
romani: Yes.
So, please remove that annotation and catch with checking exact excetion and message of it.

romani:
@vorburger , please rebase and add your Check to https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/sevntu-checks.xml for testing over our code base by Travis.

romani:
@vorburger , ping

romani:
@vorburger , ping .
Please let me know if you want to drop this PR , smb else might finish your work

romani:
@vorburger , ping .

romani:
we lost connection to Author of this code, I am closing PR .... if author or any other person ready to continue, we can reopen PR or start it from scratch.

PR #483, opened 2016-11-17
config: bump multiple versions

rnveach:
updated eclipse-cs, junit, and powermock

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ jdk:
   - oraclejdk8
 
 before_install: git clone git://git.code.sf.net/p/eclipse-cs/git eclipse-cs-git
-      && cd eclipse-cs-git/ && git checkout 6.19.1
+      && cd eclipse-cs-git/ && git checkout 7.2.0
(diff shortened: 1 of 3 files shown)


rnveach:
Even though cobertura passed, it printed lots of warnings.
https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/builds/176582451#L787
Looks like it was because of the move to Java 8.
http://stackoverflow.com/a/25382674/1016482

romani:
why StaticMethodCandidateCheckTest.java  is changed ?

rnveach:
I'll fix the formatting for the file. It was changed because it uses `import org.apache.commons.lang3.ArrayUtils;` and we removed that from the latest CS code, so it was getting a compile error.

I currently can't figure out why code coverage is saying everything is 0 on a version upgrade.

romani:
Look how upgrade to java8 was done in checkstyle, there were some changes for cobertura

rnveach:
If this passes and no more changes are requested it can be merged, I will try to work on upgrading cobertura to remove warnings.

PR #486, opened 2016-11-21
Issue #310: removed more legacy violations from sevntu

rnveach:
Issue #310 

Made some minor changes to import control for failing checks.
I believe these fixes might close out the other PR. Please note main package was changed, I believe it to be incorrect before and that lead to some confusion in the other PR.


romani:
merged! thanks a lot !!

PR #487, opened 2016-11-21
started automated testing of sonar and eclipsecs files

rnveach:
Some of the code is from checkstyle project.
Most automation is in sonar. Very similar to checkstyle's XDoc verification.
We validate almost everything except types and default values and descriptions.

Please review changes in sonar file as I am not familiar with it.


romani:
thanks a lot - impressive update.

PR #495, opened 2016-12-01
finished automated testing of sonar and eclipsecs files

rnveach:
We now validate names, classpath, properties, and messages in eclipse cs.
Quite a few mistakes, missing properties, and missing checks.

I will confirm that eclipse-cs has no issues with these changes.


romani:
thanks a lot !

Issue #463, opened 2016-07-15
ReturnCountExtendedCheck takes into account 'returns' in lambdas


baratali:
com.github.sevntu.checkstyle.checks.coding.ReturnCountExtendedCheck: Return count for 'load' method is 2 (max allowed is 1).

``` java
// code is not compilable
private static Integer load() {
    query((rs, rownum) -> {
        System.out.println();
        return p;
    });
    return a;
}
```

Similar issue for ReturnCountCheck: https://github.com/checkstyle/checkstyle/issues/1068

romani:
Fix is merged

PR #513, opened 2016-12-03
Issue #463: extended ReturnCountExtendedCheck to accept lambdas

rnveach:
Issue #463

Lambdas are now supported.
Also 100% code coverage. [Regression ](http://rveach.no-ip.org/checkstyle/regression/reports/120/) showed no NPEs.
````
<module name="ReturnCountExtendedCheck">
<property name="maxReturnCount" value="1"/>
<property name="ignoreMethodLinesCount" value="0"/>
<property name="minIgnoreReturnDepth" value="99"/>
<property name="ignoreEmptyReturns" value="false"/>
<property name="topLinesToIgnoreCount" value="0"/>
</module>
````

One thing to note, lambdas don't have a method name, so the violation message currently prints that as `null`. Should we drop that part of the message? Should we drop support for matching these patterns?

### DIFF
--- a/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.xml
+++ b/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.xml
@@ -431,6 +431,7 @@
 
       <message-key key="return.count.extended.method"/>
       <message-key key="return.count.extended.ctor"/>
+      <message-key key="return.count.extended.lambda"/>
     </rule-metadata>
 
     <rule-metadata name="%UnnecessaryParenthesesExtendedCheck.name" internal-name="UnnecessaryParenthesesExtendedCheck" parent="TreeWalker">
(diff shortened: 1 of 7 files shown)


romani:
>  Should we drop that part of the message?

yes, only for lambda.
Message should be clear to user.

> Should we drop support for matching these patterns?

I do not understand question, please rephrase the question.

rnveach:
@romani We have property `ignoreMethodsNames`, which drops methods from checking if they match.
Lambdas have no name, so can they still use this property to ignore lambdas?
Right now I convert `null` to `"null"` so you can put in the string null to the property to drop these.

romani:
ok, lets keep that hack for now, real code will show us right way to treat it.

rnveach:
@romani If we are keeping null method names for ignoring, does it make sense to still keep `null` in the message display. Users should understand that it is the method name there. If they see `null` then they may understand that they can ignore it by add `"null"` to the ignore method list.

romani:
>does it make sense to still keep null in the message display. 

No sense. We could mention this in javadoc as side effect and not a contract for behavior and could be changed in future

rnveach:
@romani Done.
Had to split violations to `logViolation` because of NestedIfDepth violation.

romani:
please rebase.

rnveach:
@romani Done.

romani:
update to functionality is big, please provide diff report.

rnveach:
Update is because regression showed the message was incorrect. I removed the method name but didn't decrease the parameter number in the property file.
We probably should create a test for something like this.

@romani Here is regression.
http://rveach.no-ip.org/checkstyle/regression/reports/128/

Issue #454, opened 2016-05-14
ConfusingConditionCheck: NPE from CS


rnveach:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (sevntu-checkstyle-check) on project checkstyle: Failed during checkstyle configuration: Exception was thrown while processing M:\checkstyleWorkspace\checkstyle\src\it\resources\com\google\checkstyle\test\chapter4formatting\rule411bracesareused\InputNeedBraces.java: NullPointerException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (sevntu-checkstyle-check) on project checkstyle: Failed during checkstyle configuration
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:216)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call(MultiThreadedBuilder.java:189)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call(MultiThreadedBuilder.java:185)
    at java.util.concurrent.FutureTask.run(FutureTask.java:262)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
    at java.util.concurrent.FutureTask.run(FutureTask.java:262)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    at java.lang.Thread.run(Thread.java:745)
Caused by: org.apache.maven.plugin.MojoExecutionException: Failed during checkstyle configuration
    at org.apache.maven.plugin.checkstyle.CheckstyleViolationCheckMojo.execute(CheckstyleViolationCheckMojo.java:543)
[...]

PR #514, opened 2016-12-03
Issue #454: fixed NPE in ConfusingConditionCheck

rnveach:
Issue #454

Fixed NPE in `getAmounOfCodeRowsInBlock`.
Since the method is counting for `blocks` I set the return to 0 if there is no block. Please let me know if we should count lines even if there is no block.

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ConfusingConditionCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ConfusingConditionCheck.java
@@ -48,7 +48,7 @@ import com.puppycrawl.tools.checkstyle.api.TokenTypes;
  * and swapped code in "if" and "else" block:
  * </p>
  * <pre>
- *  if (a == b &amp;&amp; c == d)
+ *  if (a == b || c == d)
  *      {
  *          smth2();
  *      }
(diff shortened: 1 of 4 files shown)


romani:
>Please let me know if we should count lines even if there is no block.

Please give me example of code.

rnveach:
@romani 
https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/514/commits/f8979a5d5401bc7aef7e59ed9a1e6f28df7365b3#diff-1b93271e1dfd7a8eeec046006468c167R13
````
            if(flag);
            else;
````
another example:
````
            if(flag)
very.
long.
line();
````
Code and property define use for blocks, but I assume that means between {}s and these examples have no braces.

romani:
>  It is just counting lines, not number of statements.

yes, that is all we need to do.
As screen size is certain amount of lines of code (not a statements). Idea is Check is to force user to write block code that fit a screen.

If there is no block (block is `{ ... }`) it is not a target of this Check.

romani:
Please provide:
1) stress testing report (on a lot of projects)
2) diff report (on a lot of projects, as no changes are expected)

rnveach:
@romani 
Diff report is not normally possible because this is a NPE and regression stops when an exception occurs. This is not the first time we ran into this.
I think its time we make some type of option in main checkstyle to prevent propagating the exception for these situations. I don't think it can be just an option in CLI since we are using JXR and maven-checkstyle in regression.

rnveach:
Here is stress report (**not diff**):
http://rveach.no-ip.org/checkstyle/regression/reports/130/
No NPEs.

romani:
simple launch have too much violations to state that there are no regression.

> Diff report is not normally possible because this is a NPE and regression stops when an exception occurs. 

yes this is not typical . but diff is still possible, just demand a bit more manual actions.
Please 
1) do first execution to get to know where Exception is present
2) exclude that folder or file from validation at all
....
After few launches you will get clear diff report.

-------------------------

Please rebase this branch.

rnveach:
> diff is still possible, just demand a bit more manual actions.

The point isn't it is easy to work around it, the point is the report is now false as it shows no NPEs fixed.
It is also very easy to just forge the report and hide any projects/areas that report a difference since I am cutting out areas. Some of the files I am dropping could be the one few areas were a difference could be found in all our runs.
Fix in main checkstyle should also be an easy one as to not worry about this anymore.

-------

Currently there is a difference: http://rveach.no-ip.org/checkstyle/regression/reports/134/

It happens with code like: http://rveach.no-ip.org/checkstyle/regression/reports/134/apache-ant/xref/File1.html#L152
`````
if (a) {
    ....
} else if (b) {
    ....
} else {
    ....
}
`````
My new code is saying the first else has no lines because it isn't `} else {`.
The old code is not correct in counting either. `firstBrace` is set to the second `if` instead of the brace following it, and `lastBrace` is set to the final `else` instead of the brace right before it. Depending on actual brace positions, calculation could be wrong.

rnveach:
new report on current state: http://rveach.no-ip.org/checkstyle/regression/reports/136/

http://localhost:9998/reports/136/nbia-dcm4che-tools/index.html#A1

This is a difference because of incorrect calculation of `else` block.
````
if (a) {
    ....
} else if (b) {
    ....
}
````
For the else, `firstBrace` is the second `if` statement and `lastBrace` is the brace right after the second `if` statement. It is incorrectly counting the number of lines as 0 instead of 2 in this linked case.
The new changes count the lines correct.

romani:
IF-ELSE-IF blocks can not be easily inverted , Check should skip them. Strange that is missed in javadoc.
Your last report is reasonable, please do squash.

rnveach:
@romani Squashed.

> can not be easily inverted

Why prevent them from being flagged, shouldn't it still be possible as long as second if isn't a negative?

> Check should skip them.

`IF-ELSE-IF` is not skipped.
Removals from last report are just because the ratio between the if and else (`isRatioBetweenIfAndElseBlockSuitable`) is not enough compared to `multiplyFactorForElseBlocks` to warrant a violation. Most likely there are other cases were a violation will be printed for `IF-ELSE-IF`, but this can be in a new issue.

romani:
discussion could be continued at #531

Issue #451, opened 2016-04-15
enforce commit message format and issue reference in PR description


romani:
this is already done in main project - https://github.com/checkstyle/checkstyle.

Need to be applied to this project too:
testing PR description - https://github.com/checkstyle/checkstyle/blob/master/.travis.yml#L329
testing of commit message - https://travis-ci.org/checkstyle/checkstyle/jobs/123016907#L799

rnveach:
New items added:

1)
probably as next step idea:
review https://github.com/checkstyle/sonar-checkstyle/blob/master/.travis.yml , split of sevntu travis stuff into two files will give user ability to launch easily build item on his side.

2)
we need to question why we do "javadoc:javadoc" as separate step.
"mvn clean install" should be be enough.

romani:
moved to #519 .

We should keep issues small and concise, that helps verification.

Current issue is done.

PR #518, opened 2016-12-04
Issue #451: reworked travis and added PR desc testing

rnveach:
Issue #451

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,6 @@
 language: java
 sudo: false
 
-jdk:
-  - oraclejdk8
-
-before_install: git clone git://git.code.sf.net/p/eclipse-cs/git eclipse-cs-git
-      && cd eclipse-cs-git/ && git checkout 7.2.0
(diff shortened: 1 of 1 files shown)


rnveach:
@romani What do you think of this travis rewrite?
[Before the build took 6 minutes.](https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/builds/180978673) This build took 3 minutes.
The longest being eclipse-cs because of external downloads and sadly the other items need quick install of sevntu for further checking.

If it is all ok, I will squash and merge for #451 as this will include PR test.

romani:
good !
please prepare for merge.

side note and probably as next step idea:
review https://github.com/checkstyle/sonar-checkstyle/blob/master/.travis.yml , split of travis stuff into two files will give user ability to launch easily build item on his side.

rnveach:
@romani Squashed and renamed commit.

romani:
one day we need to question why we do "javadoc:javadoc" as separate step.
"mvn clean install" should be be enough.

Issue #520, opened 2016-12-05
Sevntu-Checks: 100% code coverage


rnveach:
We should remove these suppressions, https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/4e33ce17f6f53b1fda0b212d0cd65b861baad9eb/sevntu-checks/pom.xml#L157-L192
 and aim for 100% code coverage where possible.

rnveach:
Status update:
`naming` package is only missing `EnumValueCheck` which should be done with issue #448 when it is rewritten.
`design` package has duplicated checks from main repo (`HideUtilityClassConstructorCheck`) which I don't see a point in doing. That leaves just `ChildBlockLengthCheck` and `InnerClassCheck` are left.
coding still has 14 classes left. None are duplicates.

PR #523, opened 2016-12-05
Issue #520: 100% coverage for AvoidConstantAsFirstOperandInCondition

rnveach:
Issue #520

`firstOperand.branchContains(constantType)` is removed because `constantType` gets its value from `firstOperand` at `final int constantType = firstOperand.getType();`, so it will always return true.

Regression showed no differences. http://rveach.no-ip.org/checkstyle/regression/reports/131/


romani:
please rebase to resolve merge conflict and ok to merge.

rnveach:
@romani Done.

PR #528, opened 2016-12-06
Issue #520: 100% coverage for UnnecessaryParenthesesExtendedCheck

rnveach:
Issue #520

I was able to achieve coverage for one spot, but the other 2 I couldn't achieve. Regression showed no differences when removed. http://rveach.no-ip.org/checkstyle/regression/reports/139/

` && (next.getType() == TokenTypes.RPAREN);` is not reachable because when we have the open parenthesis, we are guaranteed to have the close one.
I believe the same is true for `exprSurrounded` and we are already checking if they are surrounded else where.


romani:
When strive for 100% , it is ok to make regression, some unexpected problems are possible to appear but in long run Check become better

PR #532, opened 2016-12-06
Issue #520: 100% coverage for CauseParameterInExceptionCheck

rnveach:
Issue #520 

Added new test.


rnveach:
> [ERROR] src/test/java/com/github/sevntu/checkstyle/checks/design/CauseParameterInExceptionCheckTest.java:[146,58] (extension) ConstructorWithoutParams: Calls to constructors of 'CauseParameterInExceptionCheck' should use at least one parameter.

Am I missing something? This has no constructor parameters to use.

Edit:
>     <module name="com.github.sevntu.checkstyle.checks.design.ConstructorWithoutParams">
      <property name="classNameFormat" value=".*Exception"/>

Property in config is incorrect. It is missing `$` on end.

romani:
:) good catch.
config is changed.

PR #541, opened 2016-12-07
Issue #520: 100% coverage for ChildBlockLengthCheck

rnveach:
Issue #520 
Added tests for some cases I found in regression.

`while (curNode != null) {` and `while ((curNode != null) && `
They can't be `null` because we are in a block and are guaranteed some type of end.

Regression: http://rveach.no-ip.org/checkstyle/regression/reports/150/
No differences.


rnveach:
@romani I can't explain why it thinks this coverage is missing. I also have the problem locally.
Issue is at line 346 `if (result != 0) {`. Even testing different conditions below line proves we are meeting coverage, but that line says we are not.
Please provide guidance.

coverage

romani:
@rnveach , please rebase your changes.
I am looking into this issue now ....

rnveach:
> but I see smth different, problem is only at

Other code was just to show I am hitting coverage but that one line was acting funny.
Rebasing.

romani:
cobertura is right, there now UT for case where result == 0.
I changed  code like:
```
        if (result != 0) {
            result--;
        } else {
            throw new IllegalStateException("");
        }
```
and it works, without exception.
Simple example of dead code. We need to update code as I did and launch regression testing, exception will easily find usecase or prove that condition should be removed.

rnveach:
@romani Fixed. I probably confused myself when working on this before.
No main code was changed to get last piece of coverage. Last regression showed no differences.

PR #547, opened 2016-12-08
Issue #520: 100% coverage for SimpleAccessorNameNotationCheck

rnveach:
Issue #520

Regression: http://rveach.no-ip.org/checkstyle/regression/reports/159/

Regression for this check is almost useless because we are looking for methods with specific names. I had to disable those checks to get better results and I still couldn't appease some of the coverage spots.
I did however manage to add a lot more tests than was before.

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/SimpleAccessorNameNotationCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/SimpleAccessorNameNotationCheck.java
@@ -134,8 +134,8 @@ public class SimpleAccessorNameNotationCheck extends AbstractCheck {
             if (containsOnlyExpression(currentVerifiedTop)) {
 
                 currentVerifiedTop = currentVerifiedTop.getFirstChild();
-                final boolean containsOnlyOneAssignment = currentVerifiedTop.getChildCount() == 1
-                        && currentVerifiedTop.getFirstChild().getType() == TokenTypes.ASSIGN;
+                final boolean containsOnlyOneAssignment = currentVerifiedTop.getFirstChild()
+                        .getType() == TokenTypes.ASSIGN;
                 if (containsOnlyOneAssignment) {
 
                     currentVerifiedTop = currentVerifiedTop.getFirstChild();
@@ -207,8 +207,7 @@ public class SimpleAccessorNameNotationCheck extends AbstractCheck {
     private static boolean containsOnlyExpression(DetailAST objectBlock) {
         //three child: EXPR, SEMI and RCURLY
         return objectBlock.getChildCount() == EXPRESSION_BLOCK_CHILD_COUNT
-                && objectBlock.getFirstChild().getType() == TokenTypes.EXPR
-                && objectBlock.findFirstToken(TokenTypes.SEMI) != null;
+                && objectBlock.getFirstChild().getType() == TokenTypes.EXPR;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/SimpleAccessorNameNotationCheck.java
```diff
@@ -242,10 +241,7 @@ private static String getNameOfSettingField(DetailAST assign,
             else {
                 if (assigningFirstChild.getType() == TokenTypes.DOT) {
 
-                    if (assigningFirstChild.getChildCount() == 2
```
rnveach: Dot can only have 2 children, unless you can point me to an example that doesn't.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/SimpleAccessorNameNotationCheck.java
```diff
@@ -242,10 +241,7 @@ private static String getNameOfSettingField(DetailAST assign,
...
-                            && "this".equals(assigningFirstChild
-                                    .getFirstChild().getText())
-                            && assigningFirstChild.getLastChild().getType() == TokenTypes.IDENT) {
```
rnveach: `this` can only be followed by field name or method name, both of which are `IDENT`, right?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/SimpleAccessorNameNotationCheck.java
```diff
@@ -298,8 +294,7 @@ private static boolean containsOnlyReturn(DetailAST methodBody) {
     private static boolean isCorrectReturn(DetailAST literalReturn) {
         //two child: EXPR and SEMI
         return literalReturn.getChildCount() == 2
-                && literalReturn.getFirstChild().getType() == TokenTypes.EXPR
-                && literalReturn.getLastChild().getType() == TokenTypes.SEMI;
```
rnveach: return with expression is always guaranteed a semicolon, right? we already verify child count is 2.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/SimpleAccessorNameNotationCheck.java
```diff
@@ -324,8 +319,7 @@ private static String getNameOfGettingField(DetailAST expr) {
             else {
                 if (exprFirstChild.getType() == TokenTypes.DOT
                         && exprFirstChild.getChildCount() == 2
-                        && exprFirstChild.getFirstChild().getType() == TokenTypes.LITERAL_THIS
-                        && exprFirstChild.getLastChild().getType() == TokenTypes.IDENT) {
```
rnveach: same as removal on 248.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/SimpleAccessorNameNotationCheck.java
```diff
@@ -207,8 +207,7 @@ private boolean isGetterCorrect(DetailAST methodDef, String methodName) {
     private static boolean containsOnlyExpression(DetailAST objectBlock) {
         //three child: EXPR, SEMI and RCURLY
         return objectBlock.getChildCount() == EXPRESSION_BLOCK_CHILD_COUNT
-                && objectBlock.getFirstChild().getType() == TokenTypes.EXPR
-                && objectBlock.findFirstToken(TokenTypes.SEMI) != null;
```
rnveach: think this can't be met because we are already checking the child count.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/SimpleAccessorNameNotationCheck.java
```diff
@@ -134,8 +134,8 @@ private boolean isSetterCorrect(DetailAST methodDef, String methodName) {
             if (containsOnlyExpression(currentVerifiedTop)) {
 
                 currentVerifiedTop = currentVerifiedTop.getFirstChild();
-                final boolean containsOnlyOneAssignment = currentVerifiedTop.getChildCount() == 1
```
rnveach: probably something to do with the `containsOnlyExpression`

Issue #552, opened 2016-12-08
ForbidCertainImportsCheck: null excludes gives no violations


rnveach:
````
$ cat TestClass.java
package com.github.sevntu.checkstyle.checks.coding;

import com.puppycrawl.tools.checkstyle.api.AutomaticBean; // forbidden

public class InputForbidsCertainImports
{
    public int a()
    {
        
        AutomaticBean smth = new com.puppycrawl.tools.checkstyle.api.AutomaticBean(); // forbidden!
        return 5;
    }
}


$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
<module name="com.github.sevntu.checkstyle.checks.coding.ForbidCertainImportsCheck">
<property name="packageNameRegexp" value=".+\.sevntu\..+"/>
<property name="forbiddenImportsRegexp" value=".+\.api\..+"/>
</module>
    </module>
</module>

$ java -classpath sevntu-checks-1.22.0.jar;checkstyle-7.3-all.jar com.puppycrawl.tools.checkstyle.Main -c TestConfig.xml TestClass.java
Starting audit...
Audit done.
````

Issue only presents itself when no `forbiddenImportsExcludesRegexp` is given. Once it is set, 2 violations show up for lines 3 and 10.
It makes no sense that no exclusions would prevent any violations from showing.

romani:
fix is merged

PR #553, opened 2016-12-08
Issue #520 and #552 on ForbidCertainImportsCheck

rnveach:
Issue #520 and #552

Coverage is now 100%. Added some new tests, including one in issue #552.
`getForbiddenImportsExcludesRegexp` was removed as we never use it.
Some conditions were moved from `visit` to other method `isImportForbidden` since it is common there.
`if (parentDotAST != null) {` was removed because we either always have DOT or IDENT as first token.

`getDefaultTokens` was changed for other issue.
Only other change was `(forbiddenImportsExcludesRegexp == null  || !forbiddenImportsExcludesRegexp.matcher(importText).matches());` to make sure no NPE on a null exclude.

Regression to come.


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/161/
Config is used is the same as in the issue. All violations are new, which shows we weren't doing any validation before if ignore was not set.
Let me know if any other regression is needed.

Issue #551, opened 2016-12-08
Wiki to CI: Messages, Inputs, and JUnit


rnveach:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/wiki/What-files-need-to-be-changed-or-created-in-repository-for-new-Check-implementation

Item 2 - messages (https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/resources/com/github/sevntu/checkstyle/checks/YOUR_PACKAGE/messages.properties)

Item 3 - Junit file (https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/test/java/com/github/sevntu/checkstyle/checks/YOUR_PACKAGE/YOUR_CHECK_NAMETest.java)

Item 4 - Input files (https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/YOUR_PACKAGE/InputYOUR_CHECK_NAME.java)

The following junits should be made from these items.

1) New JUnit to verify all messages in checks exist in message properties, there is atleast 1, and can be used without any issues.
2) New JUnit to verify all checks have a test that matches its name. Extra tests will be ignored. Ex: UtilsTest and internal tests.
3) New JUnit to verify all inputs are named after the test in some fashion.
Ex: Test is named "MyCheckTest.java", inputs should match "InputMyCheck*.java".
Overrides will have to be looked at as they are found as some tests may require special file name for testing.

rnveach:
All 3 items are done.
#550 is the last Wiki item.

PR #555, opened 2016-12-08
Issue #551: verify checks have tests

rnveach:
Issue #551 - Item 2

All checks must now have test that match check name + "Test" and be located in the same package as the check is.


rnveach:
One CI node stalled. I stopped and restarted it and now it finished successfully.

PR #561, opened 2017-01-23
Issue #520: 100% coverage for EnumValueNameCheck

rnveach:
Issue #520


romani:
For some reason I thought that assert is forbidden token in checkstyle.

PR #565, opened 2017-05-09
Pull #565: added new check MoveVariableInsideIfCheck

rnveach:
Adding check `MoveVariableInsideIfCheck`.
Also added cases that can cause false positives.

Will rename commit and provide regression.

Fixes in Checkstyle: https://github.com/checkstyle/checkstyle/pull/4328
There was a small bug in my original implementation, so there may be more violations in Checkstyle than before.

### DIFF
--- a/sevntu-checks/sevntu-checks.xml
+++ b/sevntu-checks/sevntu-checks.xml
@@ -154,6 +154,7 @@
     <module name="com.github.sevntu.checkstyle.checks.coding.UselessSuperCtorCall"/>
     <module name="com.github.sevntu.checkstyle.checks.coding.UselessSingleCatch"/>
     <module name="com.github.sevntu.checkstyle.checks.coding.WhitespaceBeforeArrayInitializer"/>
+    <module name="com.github.sevntu.checkstyle.checks.coding.MoveVariableInsideIf" />
 
     <!-- Design -->
     <module name="com.github.sevntu.checkstyle.checks.design.ChildBlockLength">
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MoveVariableInsideIfCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MoveVariableInsideIfCheck.java
@@ -1,0 +1,356 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.sevntu.checkstyle.checks.coding;
+
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.ScopeUtils;
+
+/**
+ * <p>
+ * Checks if a variable is only used inside if statements and asks for it's
+ * declaration to be moved there too.
+ * </p>
+ * <p>
+ * Rationale: Code inside if/else statements are only executed when those specific block
+ * conditions evaluate to true. Moving variables inside these blocks prevents the code from being
+ * executed when the value of the variable is not even being used. It also helps limit the scope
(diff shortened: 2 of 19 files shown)


rnveach:
I changed the violation slightly so it points to the variable instead of the block so the suppression will be next to the variable, which makes more sense.

Travis failure is because of changes needed in Checkstyle.

romani:
CI has to be green.
Can you send to checkstyle repo with fixes?
Javadoc build item has to fixed too

rnveach:
Regression failed with an problem in utility, made a issue for it. Will fix it.
Will make updates to checkstyle. Will have to add suppressions for a check that doesn't exist yet.

rnveach:
Here is the regression report: http://rveach.no-ip.org/checkstyle/regression/reports/30/

Unfortunately there is another bug with the newly introduced patch only report that I did not see when making the change. I will fix that too.
The reports are still readable, it is just too many files with no violations are listed.

romani:
@rnveach , why reports are so weird and list file with no violations ?

rnveach:
@romani This is the bug I mentioned and the fix is in https://github.com/checkstyle/contribution/pull/228
I was unaware that the differentiating compare in  patch-diff-report-tool was dropping these files.

rnveach:
@romani CI passed.

sevntu-checks/pom.xml
```diff
@@ -165,6 +165,8 @@
               <regex><pattern>.*.checks.coding.EitherLogOrThrowCheck</pattern><branchRate>88</branchRate><lineRate>99</lineRate></regex>
               <regex><pattern>.*.checks.coding.ForbidThrowAnonymousExceptionsCheck</pattern><branchRate>81</branchRate><lineRate>97</lineRate></regex>
               <regex><pattern>.*.checks.coding.MapIterationInForEachLoopCheck</pattern><branchRate>90</branchRate><lineRate>98</lineRate></regex>
+              <regex><pattern>.*.checks.coding.MoveVariableInsideIfCheck</pattern><branchRate>98</branchRate><lineRate>100</lineRate></regex>
+              <regex><pattern>.*.checks.coding.MoveVariableInsideIfCheck.Holder</pattern><branchRate>89</branchRate><lineRate>100</lineRate></regex>
```
romani: can we make coverage 100% ? to completely remove this lines

sevntu-checks/sevntu-checks.xml
```diff
@@ -212,6 +212,8 @@
       <property name="ignoreMethod" value="false"/>
     </module>
 
+	<module name="com.github.sevntu.checkstyle.checks.coding.MoveVariableInsideIf" />
+
```
romani: please move it to coding group , and fix indentation

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MoveVariableInsideIfCheck.java
```diff
@@ -0,0 +1,275 @@
...
+
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.ScopeUtils;
+
+/**
+ * <p>
+ * Checks if a variable is only used inside if statements and asks for it's
+ * declaration to be moved there too.
```
romani: we do not have xdoc.
So javadoc is all that you can show to user for details. http://sevntu-checkstyle.github.io/sevntu.checkstyle/apidocs/index.html
Please update it to have wide description and examples.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MoveVariableInsideIfCheck.java
```diff
@@ -0,0 +1,275 @@
...
+     *
+     * @param holder The object holder with the specified variable to check and
+     *        it's current state.
+     * @param ifNodeGiven The current if node to examine.
+     */
+    private static void validateIf(Holder holder, DetailAST ifNodeGiven) {
+        DetailAST ifNode = ifNodeGiven;
+
+        // -@cs[SingleBreakOrContinue] Too complex to break apart
+        while (true) {
```
romani: `while TRUE` and suppression .... I do not like this.
I can not propose smth better now, but it smells.

rnveach: loop is going through all `ifs` of `if-else` chain.
while loop could be changed to make sure if node is an if node.

However, the suppression is for the number of breaks and this change won't help that.
I think it will hurt code readability by changing it, but thats just my opinion.

sevntu-checks/pom.xml
rnveach: I removed the code for 100%. Regression will tell us if there are any NPEs.

sevntu-checks/sevntu-checks.xml
rnveach: Fixed. Sorry for this.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MoveVariableInsideIfCheck.java
rnveach: Fixed.

rnveach:
New regression: http://rveach.no-ip.org/checkstyle/regression/reports/40/

No changes or NPEs.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MoveVariableInsideIfCheck.java
romani: moved to #587

romani:
now we can do 1.24.0 release

Issue #568, opened 2017-05-11
AvoidNotShortCircuitOperatorsForBooleanCheck: NPE on interface


rnveach:
````
$ cat TestClass.java
public interface TestClass {
    int ASM4 = 4 << 16 | 0;
}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
<module name="com.github.sevntu.checkstyle.checks.coding.AvoidNotShortCircuitOperatorsForBooleanCheck" />
    </module>
</module>

$ java -jar checkstyle-7.3-sevntu-1.23.0-all.jar -c TestConfig.xml TestClass.java
Starting audit...
com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing TestClass.java
    at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:287)
    at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:205)
    at com.puppycrawl.tools.checkstyle.Main.runCheckstyle(Main.java:425)
    at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:359)
    at com.puppycrawl.tools.checkstyle.Main.main(Main.java:174)
Caused by: java.lang.NullPointerException
    at com.github.sevntu.checkstyle.checks.coding.AvoidNotShortCircuitOperatorsForBooleanCheck.isBooleanExpression(AvoidNotShortCircuitOperatorsForBooleanCheck.java:145)
    at com.github.sevntu.checkstyle.checks.coding.AvoidNotShortCircuitOperatorsForBooleanCheck.visitToken(AvoidNotShortCircuitOperatorsForBooleanCheck.java:112)
    at com.puppycrawl.tools.checkstyle.TreeWalker.notifyVisit(TreeWalker.java:361)
    at com.puppycrawl.tools.checkstyle.TreeWalker.processIter(TreeWalker.java:498)
    at com.puppycrawl.tools.checkstyle.TreeWalker.walk(TreeWalker.java:303)
    at com.puppycrawl.tools.checkstyle.TreeWalker.processFiltered(TreeWalker.java:178)
[...]

PR #569, opened 2017-05-12
Issue #568: fixed NPE in non-classes for ShortCircuitOperators check

rnveach:
Issue #568

Regression to come.


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/38/

Issue #427, opened 2016-01-25
MapIterationInForEachLoopCheck crashes on classes extending Maps


karlicoss:
I managed to reproduce it in a unit test https://github.com/sevntu-checkstyle/sevntu.checkstyle/commit/0587683c168259882b02adb29a377def7cae5b3f, but not really sure how it should be fixed.

Stacktrace:

```
java.lang.NullPointerException
    at com.github.sevntu.checkstyle.checks.coding.MapIterationInForEachLoopCheck.getKeySetOrEntrySetNode(MapIterationInForEachLoopCheck.java:400)
    at com.github.sevntu.checkstyle.checks.coding.MapIterationInForEachLoopCheck.validate(MapIterationInForEachLoopCheck.java:332)
    at com.github.sevntu.checkstyle.checks.coding.MapIterationInForEachLoopCheck.visitToken(MapIterationInForEachLoopCheck.java:306)
    at com.puppycrawl.tools.checkstyle.TreeWalker.notifyVisit(TreeWalker.java:388)
    at com.puppycrawl.tools.checkstyle.TreeWalker.processIter(TreeWalker.java:499)
    at com.puppycrawl.tools.checkstyle.TreeWalker.walk(TreeWalker.java:330)
    at com.puppycrawl.tools.checkstyle.TreeWalker.processFiltered(TreeWalker.java:203)
    at com.puppycrawl.tools.checkstyle.api.AbstractFileSetCheck.process(AbstractFileSetCheck.java:79)
    at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:265)
    at com.github.sevntu.checkstyle.BaseCheckTestSupport.verify(BaseCheckTestSupport.java:83)
    at com.github.sevntu.checkstyle.BaseCheckTestSupport.verify(BaseCheckTestSupport.java:75)
    at com.github.sevntu.checkstyle.BaseCheckTestSupport.verify(BaseCheckTestSupport.java:64)
    at com.github.sevntu.checkstyle.checks.coding.MapIterationInForEachLoopTest.extendsMapTest(MapIterationInForEachLoopTest.java:88)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #570, opened 2017-05-12
Issue #427: fixed NPE in MapIterationInForEachLoopCheck for map classes

rnveach:
Issue #427

Regression to come.


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/41/

Only exceptions removed.

PR #574, opened 2017-05-12
Issue #573: fixed NPE in NoNullForCollectionReturnCheck for constructors

rnveach:
Issue #573

Regression to come.


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/47/

Some exceptions turned into violations.

Issue #580, opened 2017-05-14
MapIterationInForEachLoopCheck: NPE in class that extends map


rnveach:
identified at https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/427#issuecomment-301251881 but left undone.

````
$ cat TestClass.java
package com.github.sevntu.checkstyle.checks.coding;

import java.util.HashMap;

public class InputMapIterationInForEachLoopExtends {
    public static class TestMap extends HashMap<Integer, Integer> {
        public void test() {
            for (Entry<Integer, Integer> entry : entrySet()) {
            }
        }
    }
}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
<module name="com.github.sevntu.checkstyle.checks.coding.MapIterationInForEachLoopCheck">
  <property name="proposeValuesUsage" value="true"/>
  <property name="proposeKeySetUsage" value="true"/>
  <property name="proposeEntrySetUsage" value="true"/>
</module>
    </module>
</module>

$ java -jar checkstyle-7.3-sevntu-1.23.0-all.jar -c TestConfig.xml TestClass.java
Starting audit...
com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing TestClass.java
    at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:287)
    at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:205)
    at com.puppycrawl.tools.checkstyle.Main.runCheckstyle(Main.java:425)
    at com.puppycrawl.tools.checkstyle.Main.runCli(Main.java:359)
    at com.puppycrawl.tools.checkstyle.Main.main(Main.java:174)
Caused by: java.lang.NullPointerException
    at com.github.sevntu.checkstyle.checks.coding.MapIterationInForEachLoopCheck.getKeySetOrEntrySetNode(MapIterationInForEachLoopCheck.java:406)
    at com.github.sevntu.checkstyle.checks.coding.MapIterationInForEachLoopCheck.validate(MapIterationInForEachLoopCheck.java:332)
[...]

PR #583, opened 2017-05-14
Issue #580: fixed NPE in MapIterationInForEachLoop for maps without this

rnveach:
Issue #580

Regression to come.


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/50/

PR #584, opened 2017-05-14
Issue #578: fixed NPE in FinalizeImplementationCheck

rnveach:
Issue #578

Regression to come.


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/51/

PR #585, opened 2017-05-15
Issue #577: fixed NPE in AvoidHidingCauseException on wrapping

rnveach:
Issue #577

Regression to come.


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/52/

Issue #448, opened 2016-04-04
redo EnumValueCheck: to make it simple as all other Naming Checks


romani:
reason of bad original design: https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/370#issue-99001539

reason to keep it simple: https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/370#issuecomment-199290433

romani:
@pbaranchikov , can you help us to implement this issue ?

rnveach:
We should achieve 100% code coverage with this rewrite.

rnveach:
@romani `AbstractNameCheck` only takes 1 regular expression.
`EnumValueNameCheck` uses 2 expressions for what looks like if the `enum` is defined inside a class or is defined in a file all by itself.

Do you expect 2 checks out of this re-implementation?

romani:
I thought we decided to not differentiate enums. So one regexp should stay

PR #586, opened 2017-05-15
Issue #448: rewrote EnumValueNameCheck to follow naming check pattern

rnveach:
Issue #448

This may cause issues on CI regression to main project since we are removing properties in the rewrite and checkstyle is already using it.


rnveach:
> Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Property 'objFormat' in module EnumValueNameCheck does not exist, please check the documentation

We will have to fix checkstyle when it upgrades sevntu.
This should only be merged once we are closer to a sevntu release.

romani:
As we finish PR with new Check that you introduce I will merge this PR

rnveach:
@romani Don't forget this.

Issue #589, opened 2017-05-26
CauseParameterInExceptionCheck: ArrayIndexOutOfBoundsException


rnveach:
Identified at https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/571#issue-228381444 , and was able to reproduce with the latest version of checkstyle in report http://rveach.no-ip.org/checkstyle/regression/reports/61/apache-ant/index.html#A4

````
Got an exception - java.lang.ArrayIndexOutOfBoundsException: 586
at com.puppycrawl.tools.checkstyle.api.AbstractCheck.log(AbstractCheck.java:253)
at com.puppycrawl.tools.checkstyle.api.AbstractViolationReporter.log(AbstractViolationReporter.java:144)
at com.github.sevntu.checkstyle.checks.design.CauseParameterInExceptionCheck.finishTree(CauseParameterInExceptionCheck.java:174)
````

Exception is happening because [the field `exceptionClassesToWarn` isn't being cleared when processing multiple files when an exception is thrown on the previous file's run](https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/f24223e0eeb9aec44055c6a56e16386ddd5a6a12/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/CauseParameterInExceptionCheck.java#L176). The problem is only surfacing because of `haltOnException` being off.
`finishTree` shouldn't be used to reset instance fields. It is not guaranteed to always be called, like on an exception.

Here is the source of the `checkstyle-result.xml`:
````
<file name="/home/ricky/opensource/contribution/checkstyle-tester/reports/saverefs/apache-ant/src/main/org/apache/tools/ant/taskdefs/optional/ejb/IPlanetEjbc.java">
[...]

PR #590, opened 2017-05-26
Issue #589: fixed OutOfBoundsException in CauseParameterInException

rnveach:
Issue #589

Moved field reset to `beginTree`.


rnveach:
Travis failure is because checkstyle hasn't been updated for new EnumValueNameCheck.

I looked over all sevntu checks, and didn't find any other occurrences of a possible `finishTree` issue like this.

PR #592, opened 2017-05-31
config: allowed overriding checkstyle configuration for snapshots

rnveach:
These are changes needed for no error testing from main checkstyle project as talked about at https://github.com/checkstyle/checkstyle/issues/4387#issuecomment-305205047 .

We could already override the checkstyle version number, but the configuration is based on that version number. If CS version is switched to a SNAPSHOT, it will try to load that from github but SNAPSHOT configurations don't exist there. Instead with new changes, wercker will override and give it the PR's configuration location.

One Question, will this be no exception testing (not using `launch.groovy`) or no error testing?
No error testing would be more thorough, and thats what it should be with current changes.
No exception testing would require us to be able to override `failOnViolation` to turn it off, which would be another change in this PR.

### DIFF
--- a/sevntu-checks/pom.xml
+++ b/sevntu-checks/pom.xml
@@ -17,6 +17,7 @@
     <project.build.sourceEncoding>iso-8859-1</project.build.sourceEncoding>
     <!-- it is not compile dependency to checkstyle, this version is only for maven plugins -->
     <checkstyle.version>7.8</checkstyle.version>
+    <checkstyle.configLocation>https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-${checkstyle.version}/config/checkstyle_checks.xml</checkstyle.configLocation>
     <!-- It is compile dependency to checkstyle, this version has to be the same as eclispe-cs depends on -->
     <checkstyle.eclipse-cs.version>7.6</checkstyle.eclipse-cs.version>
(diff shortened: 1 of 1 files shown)


rnveach:
CI failure is just key problem.
> Caused by: java.lang.IllegalStateException: Duplicate key

romani:
> One Question, will this be no exception testing (not using launch.groovy) or no error testing?

this case is really hard to catch. We just got lucky that we were lazy to remove `com.github.sevntu.checkstyle.checks.design.HideUtilityClassConstructorCheck` Check from sevntu.
One day we will do this.
execution of launch.groovy or ..... will not do any valuable testing.
It might be reasonable by some CI to:
1) build snapshot checkstyle 
2) clone https://github.com/sevntu-checkstyle/checkstyle-samples
3) update pom file to use snapshot or do this by variable
4) just launch maven example only, to make sure samples works.

rnveach:
> It might be reasonable by some CI to:

I was going to do all this in Checkstyle repo but item 2 was going to be a clone of sevntu, hence this PR fix.
The question was should the CI be a no error test (regression must show no new violations) or can we show new violations and just fail on exceptions.

romani:
@rnveach , lets make a CI execution point that check certain violation check, if that is complicated - do "not exception" validation.

PR #600, opened 2017-07-19
Pull #: removed excess hierarchy from NestedSwitchCheck

rnveach:
Fix for checkstyle/checkstyle#4767 .
We are removing deprecated classes in next release, so we need this class to extend `AbstractCheck` directly.

Will rename commit once I get a PR number.


rnveach:
Tests passed. I am merging.

Issue #525, opened 2016-12-05
Resolve sevntu-checks violations in its own code (old code)


romani:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/suppressions.xml
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/sevntu-suppressions.xml

There are suppressions like "START of legacy code, ...."
All of them should be resolved.

pbludov:
One more PMD rule to resolve: https://pmd.github.io/pmd-6.35.0/pmd_rules_java_design.html#cognitivecomplexity
```
[INFO] PMD Failure: com.github.sevntu.checkstyle.checks.coding.AvoidHidingCauseExceptionCheck:200 Rule:CognitiveComplexity Priority:3 The method 'makeExceptionsList(DetailAST, DetailAST, String)' has a cognitive complexity of 24, current threshold is 15.
[INFO] PMD Failure: com.github.sevntu.checkstyle.checks.coding.CustomDeclarationOrderCheck:368 Rule:CognitiveComplexity Priority:3 The method 'visitToken(DetailAST)' has a cognitive complexity of 39, current threshold is 15.
[INFO] PMD Failure: com.github.sevntu.checkstyle.checks.coding.CustomDeclarationOrderCheck:625 Rule:CognitiveComplexity Priority:3 The method 'getPositionInOrderDeclaration(DetailAST)' has a cognitive complexity of 19, current threshold is 15.
[INFO] PMD Failure: com.github.sevntu.checkstyle.checks.coding.CustomDeclarationOrderCheck$ClassDetail:1381 Rule:CognitiveComplexity Priority:3 The method 'getWrongOrderedGettersSetters(List, int)' has a cognitive complexity of 18, current threshold is 15.
[INFO] PMD Failure: com.github.sevntu.checkstyle.checks.coding.EitherLogOrThrowCheck:360 Rule:CognitiveComplexity Priority:3 The method 'processCatchNode(DetailAST)' has a cognitive complexity of 25, current threshold is 15.
[INFO] PMD Failure: com.github.sevntu.checkstyle.checks.coding.ForbidCertainMethodCheck$IntRange:255 Rule:UnusedPrivateMethod Priority:3 Avoid unused private methods such as 'from(String)'..
[...]

PR #609, opened 2017-09-28
Issue #525: removed return count suppression and fixed violations

rnveach:
Issue #525


rnveach:
Travis failure is still eclipse-cs and weird issue with PR reference test.

PR #611, opened 2017-10-02
Issue #525: removed javadoc package suppression and fixed violations

rnveach:
Issue #525


rnveach:
travis failure is still weird issue with testing issue reference.

PR #615, opened 2017-10-23
Issue #525: removed modified control suppression and fixed violations

rnveach:
Issue #525


romani:
some code is removed  from Check ... strange  that UTs did not catch that.

@rnveach , I do believe that regression is required here.

rnveach:
@romani Here is regression: http://rveach.no-ip.org/checkstyle/regression/reports/130/
No differences.

> strange that UTs did not catch that.

UTs wouldn't have caught this except checkstyle and pitest.
Even though we are modifying the loop variable, the for-each strictly controls the loop through it's iterator. So whatever we assign it is overwritten in the next go through.
Even if the loop variable re-assignment was kept, it still has no effect because we are looking for a specific token, which may appear only every other token like a comma separated list.

Issue #618, opened 2017-10-26
New check: to ensure all try/catch in tests have a fail in the try block


rnveach:
Based on issue at https://github.com/checkstyle/checkstyle/issues/5210 , this issue is to produce a new check to automatically catch these cases.

romani:
fix is merged.

PR #619, opened 2017-10-26
Issue #618: added RequireFailForTryCatchInJunitCheck

rnveach:
Issue #618

I expect this to fail as changes need to be made to the main project and I need to update the XML files for this check.

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
@@ -1,0 +1,246 @@
+////////////////////////////////////////////////////////////////////////////////
+// checkstyle: Checks Java source code for adherence to a set of rules.
+// Copyright (C) 2001-2017 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+////////////////////////////////////////////////////////////////////////////////
+
+package com.github.sevntu.checkstyle.checks.coding;
+
+import com.github.sevntu.checkstyle.Utils;
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.FullIdent;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+import com.puppycrawl.tools.checkstyle.utils.AnnotationUtility;
+
+/**
+ * <p>
+ * Checks if a try/catch block has a junit fail assertion inside the try for a junit method.
+ * </p>
+ * <p>
+ * Rationale: Tests should not complete the try block naturally if they are expecting a failure.
+ * If the try completes normally the test will pass successfully and skip over any assertions in
+ * the catch block.
+ * If tests are not expecting exceptions, then they should remove the catch block and propagate
+ * the exception to the junit caller which will display the full exception to the user.
+ * </p>
+ * <p>
+ * A junit test method is identified by the annotations placed on it. It is only considered a junit
+ * method if it contains the annotation 'org.junit.Test'. This check doesn't examine methods called
+ * by a test method. It must contain the annotation. Failures are identified by the
+ * method call to the method 'org.junit.Assert.fail'.
+ * </p>
+ * <p>
+ * An example of how to configure the check is:
+ * </p>
+ * <pre>
+ * &lt;module name="RequireFailForTryCatchInJunitCheck"/&gt;
+ * </pre>
+ * <p>
+ * which will cause a violation in the example below:
+ * </p>
+ * <pre>
+ *   &#064;Test
+ *   public void testMyCase() {
+ *     try { // violation here as try block has no 'Assert.fail()'.
+ *       verifySomeResult();
+ *     }
+ *     catch (IllegalArgumentException ex) {
+ *       assertEquals("expected exception message",
+ *           "Some message that is expected", ex.getMessage());
+ *     }
+ *   }
+ * </pre>
+ * @author Richard Veach
+ */
+public class RequireFailForTryCatchInJunitCheck extends AbstractCheck {
+    /**
+     * Violation message key.
+     */
+    public static final String MSG_KEY = "require.fail";
+
+    /**
+     * Fully qualified junit test annotation.
+     */
+    private static final String FQ_JUNIT_TEST = "org.junit.Test";
+    /**
+     * JUnit's fail assertion method name.
+     */
+    private static final String FAIL = "fail";
+
+    /**
+     * {@code true} if the junit test is imported.
+     */
+    private boolean importTest;
+    /**
+     * {@code true} if the junit assert is imported.
+     */
+    private boolean importAssert;
+    /**
+     * {@code true} if the junit fail assertion method is statically imported.
+     */
+    private boolean importStaticFail;
+
+    @Override
+    public int[] getDefaultTokens() {
+        return getRequiredTokens();
+    }
+
+    @Override
+    public int[] getAcceptableTokens() {
+        return getRequiredTokens();
+    }
+
+    @Override
+    public int[] getRequiredTokens() {
+        return new int[] {
+            TokenTypes.IMPORT,
+            TokenTypes.STATIC_IMPORT,
+            TokenTypes.LITERAL_TRY,
+        };
+    }
+
+    @Override
+    public void beginTree(DetailAST rootAST) {
+        importTest = false;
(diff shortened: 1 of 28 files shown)


rnveach:
Fixed all test errors and updated the check to ignore the situation found at https://github.com/checkstyle/checkstyle/issues/5210#issuecomment-339812798 .

romani:
travis is restarted

rnveach:
@romani All CIs passed. Please review.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
```diff
@@ -0,0 +1,215 @@
...
+                break;
+        }
+    }
+
+    /**
+     * Examines the try block for violations.
+     * @param ast The try block to examine.
+     */
+    private void examineTry(DetailAST ast) {
+        final DetailAST method = getStopParent(ast);
```
romani: > getStopParent

hard to understand....it is better to name it `getMethod()` as you really care only about methods. In case of lambda this method should return null.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
```diff
@@ -0,0 +1,215 @@
...
+
+        if (method != null
+                && method.getType() == TokenTypes.METHOD_DEF
+                && ast.findFirstToken(TokenTypes.LITERAL_CATCH) != null
```
romani: move "ast.findFirstToken(TokenTypes.LITERAL_CATCH) != null" to inner IF.

and you you see that all expessions that are left could be named as `isTestMethod`, logic will become to simple to read.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
rnveach: @romani I originally had it as `getMethod` but then I thought the addition of lambdas broke that as the name doesn't warn about lambdas. I couldn't think of a good name, hence the stop parent.

I will rename this to `getMethod` again.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
```diff
@@ -0,0 +1,215 @@
...
+                Utils.reportInvalidToken(ast.getType());
+                break;
+        }
+    }
+
+    /**
+     * Examines the try block for violations.
+     * @param ast The try block to examine.
+     */
+    private void examineTry(DetailAST ast) {
```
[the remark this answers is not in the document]
rnveach: @romani Is this a new requirement we start using this utility? Will we be applying it to all checks?
Do you want to start an issue on it?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
rnveach: Done.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
rnveach: Done.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
```diff
@@ -0,0 +1,215 @@
...
+ * Checks if a try/catch block has a junit fail assertion inside the try  for a junit method.
+ * </p>
+ * <p>
+ * Rationale: Tests should not complete the try block naturally if they are expecting a failure.
+ * If the try completes normally the test will pass successfully and skip over any assertions in
+ * the catch block.
+ * If tests are not expecting exceptions, then they should remove the catch block and propagate
+ * the exception to the junit caller which will display the full exception to the user.
+ * </p>
+ * <p>
```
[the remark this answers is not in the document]
rnveach: We don't identify test class, just test method. We won't examine methods called by a test method.

Done.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
rnveach: `method.getType() == TokenTypes.METHOD_DEF` was removed because of the change to `getMethod` which returns null as long as non-methods aren't returned.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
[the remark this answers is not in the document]
rnveach: @romani Please explain how to read this report.
http://rveach.no-ip.org/reports/RequireFailForTryCatchInJunitCheck.java.html

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
```diff
@@ -0,0 +1,215 @@
...
+ * An example of how to configure the check is:
+ * </p>
+ * <pre>
+ * &lt;module name="RequireFailForTryCatchInJunitCheck"/&gt;
+ * </pre>
```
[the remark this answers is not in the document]
rnveach: Done.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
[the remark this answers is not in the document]
romani: >Is this a new requirement we start using this utility?

Not a mandatory, but I ask to follow when I notice that problem in code

>Will we be applying it to all checks?Do you want to start an issue on it?

It is in plan for GSoC.

>Please explain how to read this report 

It is good sign that we did documentation in bad way, I will appreciate your changes do doc  ones all become clear

We need to make all red cell values on right side of diagonal - to have declaration of method after it usage.
Minimize distance between us and declaration to let human read code by scrolling from to to bottom, without jumps.
getImportText(DetailAST) - place it to the end of file, as just to it is unavoidable so distance of jump does not matter that much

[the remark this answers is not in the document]
rnveach: > getImportText(DetailAST) - place it to the end of file

Done.

> We need to make all red cell values on right side of diagonal
> getImportText(DetailAST) - place it to the end of file

Why move this one when it is already on the right side of the diagonal?

I assume the main goal is to remove all the red and have perfect alignment.
What is the secondary goal, having the top numbers total to the smallest number possible?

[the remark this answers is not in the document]
romani: moving to right of diagonal is first step to place all declarations after usages.

next step is to optimize distances. Main reason of this metric is to minimize jumps(navigation to far part of code) of user over the code. But if jump(long distance) is unavoidable or let other methods be close to each other. So if you move method getImportText to the end, all other should become more liner in sequence.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
```diff
@@ -0,0 +1,246 @@
...
+        return getDefaultTokens();
+    }
+
+    @Override
+    public int[] getDefaultTokens() {
+        return new int[] {
+            TokenTypes.IMPORT,
+            TokenTypes.STATIC_IMPORT,
+            TokenTypes.LITERAL_TRY,
+        };
```
romani: declaration of such tokens should be at getRequiredTokens.

rnveach: Done.

romani:
@rnveach , should we start pitest coverage demand ?

if Check is going to go to main project , that will be mandatory. Resolving pitest missed coverage after some time  could be over complicated. If Check might stay in this project forever ... pitest is not requirement as this project is for testing of Check in real life code before finalization in design, so demands for quiality could be a bit less severe.

rnveach:
@romani If sevntu should be less strict then main project, then enforcing pitest is not the way to go. You even said yourself that pitest was getting very complicated for some of the last checks not at 100%. I am more concerned with quality in main project then here unless more people help out with maintaining this side or it becomes more active.

If we did enable it for fun here and not demanding a strict percentage, I feel this would just push back on the reviewers to help the coder get a higher percentage for new code. I'm not sure how many people would understand what pitest is and how to achieve it. I know subbu in main project was confused by pitest when I tried to explain it to him.

romani:
> And how do we know if a check will stay or not?

if Check can run on random code without masive amount of false positives/want-fix it should go to main project. If usage if Check is specific to some conner cases and should be havily used with supporessions, it should probably stay in sevntu.

romani:
> You even said yourself that pitest was getting very complicated for some of the last checks not at 100%.

it is much easy to reach 100% on moment of writing, rather than in several month, when full context of code was forgotten.

PR #605, opened 2017-09-13
Issue #604: Add a check to ban use of specified method

raghavgautam:
Issue #604

PR #624, opened 2017-11-26
minor: test missing method

rnveach:
Currently testing out missing method that needs to be included.

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/CheckstyleTestMakeupCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/CheckstyleTestMakeupCheck.java
@@ -98,7 +98,7 @@ public class CheckstyleTestMakeupCheck extends AbstractCheck {
 
     /** Regular expression for matching a create method by name. */
     private Pattern createMethodRegexp = Pattern
-            .compile("create(Root|Module)Config|getModuleConfig");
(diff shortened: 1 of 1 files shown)


rnveach:
Regression says no new violations. I will keep this open until I'm sure I didn't miss anything else.

rnveach:
based on what I saw in https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/605 , I may need to expand the check to flag methods that aren't `@Test` but call `verify` with no try/catch.
These type of methods are hidden from regression's view.

rnveach:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/CheckstyleTestMakeupCheck.java#L220
We should update check to make sure we can identify class being created by `create` method since that is what we will be pulling in to identify for regression.

romani:
@rnveach , do you need this PR ?

rnveach:
@romani I am currently using it as a reminder of things I need to do for the regression check and project.
I plan to come back to this soon. I am waiting on finalizing the check in the regression project before I finalize everything that needs to be updated in this check.

romani:
@rnveach , is there any plans to come back to this PR ? It doesn't like reminder that it actually works.

romani:
@rnveach , I am closing this PR, to keep list of PRs clean.

PR #628, opened 2018-01-03
Upgrade Sevntu to Checkstyle 8.5

rnveach:
We broke compatibility between 8.6 and 8.7 such that we can't intermingle versions of sevntu with those checkstyle versions.
Eclipse-cs ingrains checkstyle jar inside itself and it is stuck on 8.5 . Since sevntu's eclipse plugin must use the sevntu jar, sevntu must be compiled with the same checkstyle version. If it is not, eclipse-cs users will see the same errors we do now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,14 +60,15 @@ matrix:
         - CMD=$CMD0$CMD1$CMD2$CMD3
 
     # regression on checkstyle
-    - jdk: oraclejdk8
-      env:
-        - DESC="checkstyle regression"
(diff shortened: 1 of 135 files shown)


rnveach:
Checkstyle regression must be disabled because we must compile code on 8.5 but use 8.8-SNAPSHOT on regression. I am worried that we will hit a point where the 2 can't work together easily.
We can't use checkstyle 8.5 in regression as we would never be able to get around violations that would be fixed in 8.8-SNAPSHOT code.

Issue #630, opened 2018-01-03
Remove custom sevntu test code and re-use checkstyle's test code


rnveach:
remove `BaseCheckTestSupport` and import `AbstractModuleTestSupport` directly from checkstyle.
remove `CheckUtil` and make `SevntuCheckUtil` that makes calls to either Checkstyle's `CheckUtil` or `ModuleReflectionUtils`.

`CheckUtil` in checkstyle may need some changes so it can be extended properly.

romani:
BaseCheckTestSupport is removed.

PR #631, opened 2018-01-03
Issue #630: use AbstractModuleTestSupport directly from checkstyle

rnveach:
Issue #630

First of probably 3 PRs.
We remove `BaseCheckTestSupport` for `AbstractModuleTestSupport`.


rnveach:
Travis failure is because issue isn't approved.

PR #652, opened 2018-03-28
minor: follow redirect

jsoref:
`$ curl -v https://checkstyle.github.io/eclipse-cs/dtds/checkstyle-metadata_1_1.dtd 2>&1|grep Loc`

```
< Location: http://checkstyle.org/eclipse-cs/dtds/checkstyle-metadata_1_1.dtd
```

### DIFF
--- a/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/annotation/checkstyle-metadata.xml
+++ b/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/annotation/checkstyle-metadata.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE checkstyle-metadata PUBLIC
 "-//eclipse-cs//DTD Check Metadata 1.1//EN"
-"https://checkstyle.github.io/eclipse-cs/dtds/checkstyle-metadata_1_1.dtd">
+"http://checkstyle.org/eclipse-cs/dtds/checkstyle-metadata_1_1.dtd">
 <checkstyle-metadata>
   <rule-group-metadata name="SevNTU checks" priority="500">
 
--- a/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.xml
+++ b/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE checkstyle-metadata PUBLIC
 "-//eclipse-cs//DTD Check Metadata 1.1//EN"
-"https://checkstyle.github.io/eclipse-cs/dtds/checkstyle-metadata_1_1.dtd">
+"http://checkstyle.org/eclipse-cs/dtds/checkstyle-metadata_1_1.dtd">
 <checkstyle-metadata>
   <rule-group-metadata name="SevNTU checks" priority="1600">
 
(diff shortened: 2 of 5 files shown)


eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/annotation/checkstyle-metadata.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE checkstyle-metadata PUBLIC
 "-//eclipse-cs//DTD Check Metadata 1.1//EN"
-"https://checkstyle.github.io/eclipse-cs/dtds/checkstyle-metadata_1_1.dtd">
+"http://checkstyle.org/eclipse-cs/dtds/checkstyle-metadata_1_1.dtd">
```
rnveach: Shouldn't we stick with `https`?

jsoref: In theory, yes.
In practice:
1. your github url redirects to http
2. your web server doesn't have a proper certificate:
> Your connection is not private
> Attackers might be trying to steal your information from checkstyle.org (for example, passwords, messages, or credit cards). Learn more
> NET::ERR_CERT_COMMON_NAME_INVALID
>  
> Automatically send some system information and page content to Google to help detect dangerous apps and sites. Privacy policy

jsoref: To be clear. I'm a huge fan of TLS certificates (I contribute actual code to letsencrypt).
But, as a drive-by-contributor, if my contribution enables you to recognize a flaw in your system, that shouldn't be my problem or impact my contribution.

As it is, today, your code includes an `https:` marker which tricks people into thinking they're getting content securely, in fact, they aren't, because whatever UA they're using is going to follow that redirect to the insecure site (and possibly across a MITM if there is one).

You and your users would be better served by having a clearly visible `http:` url to highlight the fact that it's insecure, while someone else works on getting a proper TLS certificate set up for the domain. And once that's fixed, someone else can fix all `http:` urls (not just these) to use `https:`.

rnveach: @romani ping

romani: we had related conversation about https usage at checkstyle/checkstyle#5601 .
We are not ready right now to use https. I would rather avoid this as much as possible or smb educate us how to get certificate for free and .... .

there was "http" originally -  https://github.com/sevntu-checkstyle/sevntu.checkstyle/commit/09367f08cea4615a2f0ea3ed4125c79244c2e481 , it was my bad to use "https" as I just followed redirections when eclipse-cs migrated to github.io (which use https by default).

I am ok to use http for now. Ones we finally migrate to github.io we can think about https usage.

daniilyar: @romani that is easy to get a free certificate from e.g. letsencrypt (letsencrypt even supported wildcard certificates recently), but unfortunately GitHub can't be configured with that certificate. 

If we want to have a valid TLS cert together with custom domain names (*.checkstyle.org) then we need to host some Nginx in the cloud and put the cert to it so that proxy will do SSL termination and http --> https redirects for us.

Also as all free TLS certs on the market are short-living ones that require refresh each few months. So we will also need to setup the certificate auto-renew in this case

PR #651, opened 2018-03-28
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

PR #655, opened 2018-04-02
spelling: complicated

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

split from #651


sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/CustomDeclarationOrderCheck.java
```diff
@@ -203,7 +203,7 @@
 
     private static final String METHOD_MACRO = "Method";
 
-    private static final String ANNON_CLASS_FIELD_MACRO = "DeclareAnnonClassField";
+    private static final String ANON_CLASS_FIELD_MACRO = "DeclareAnonClassField";
```
rnveach: > Caused by: java.lang.IllegalArgumentException: Unable to parse DeclareAnnonClassField

Regression is failing because of this change.
This is a breaking change for this check.

romani: I am ok with change, but please move this to special issue, it should be done visibly to users and plugins and ....

jsoref:
@rnveach:  yes

rnveach:
@romani Please review.

Issue #662, opened 2018-05-27
Update all AbstractChecks to log DetailAST


rnveach:
All AbstractChecks should use log(DetailAST ast, String key, Object... args) method to log DetailAST.

Similar to https://github.com/checkstyle/checkstyle/issues/4830

rnveach:
The only checks left should be `ForbidCCCommentsInMethods` (whole check has to be rewritten) and `LineLengthExtended` (not sure how to convert it over).

romani:
> ForbidCCCommentsInMethods

we can put violation on any AST that exists before comment.

> LineLengthExtended

we can put violation on first AST in line.

rnveach:
> put violation on any AST that exists before comment
> put violation on first AST in line

Obviously, but finding these ASTs is the problem. We have no util that finds AST right before line or on line, etc... Even if we did have such a utility, it would almost require to reparse a chunk of the tree.

rnveach:
`MultipleStringLiteralsExtendedCheck` also needs to be converted. I confirmed that these are the only 3 left.

rnveach:
@romani Once `ForbidCCCommentsInMethods` is merged, I feel we should just move `LineLengthExtendedCheck` to a new issue and close this one.

There is no easy way to convert it. The check is comment aware and empty line aware. It calls `getLines()` and can see everything in the file. To me this check seems like it is an example of a `FileSet` check with a `AbstractCheck` that acts as a holder. I know we don't like holders, but I don't see any other good way.

romani:
> to a new issue and close this one.

agree.

> I know we don't like holders, but I don't see any other good way.

lets keep it as is, if smb care about it, it will be updated. There will be more intensive to update it.

PR #667, opened 2018-05-27
Issue #662: moved complicated line Checks to log AST (part 1)

rnveach:
Issue #662

Regression should be run on either log or throw since it has a possibility for NPE.


rnveach:
http://rveach.no-ip.org/checkstyle/regression/reports/157/

No exceptions. Only see column number change differences.

Issue #669, opened 2018-05-27
Remove usage of DetailAST.branchContains


rnveach:
Version of https://github.com/checkstyle/checkstyle/issues/5124 for sevntu checks

rnveach:
4 checks still have the method
https://github.com/sevntu-checkstyle/sevntu.checkstyle/search?q=branchContains&unscoped_q=branchContains

nmancus1:
We can remove suppressions from https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/842 when this is closed.

PR #670, opened 2018-05-27
Issue #669: removed branchContains for modifiers

rnveach:
Issue #669

Regression to come.


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/158/

No differences.

PR #676, opened 2018-06-07
Issue #662: moved ForbidCCommentsInMethodsCheck to log AST

rnveach:
Issue #662 

I rewrote the check to use `visitToken` and log on the AST instead of relying on `getFileContents()`.
Tests show no differences. I will confirm the same with regression.

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ForbidCCommentsInMethodsCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ForbidCCommentsInMethodsCheck.java
@@ -19,7 +19,8 @@
 
 package com.github.sevntu.checkstyle.checks.coding;
 
-import java.util.Set;
+import java.util.ArrayDeque;
+import java.util.Deque;
 
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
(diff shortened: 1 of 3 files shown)


sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ForbidCCommentsInMethodsCheck.java
```diff
@@ -39,15 +40,21 @@
      */
     public static final String MSG_KEY = "forbid.c.comments.in.the.method.body";
 
-    /**
-     * Set contains C style comments from current file.
-     */
-    private Set<Integer> clangComments;
+    /** AST to use for null checks because {@link Deque} doesn't allow null elements. */
+    private static final DetailAST NULL_AST = new DetailAST();
```
rnveach: `Deque` does not allow `null` elements to be placed inside it.
I originally wanted a Boolean to say in method or not, but I need to keep track of the current method AST for checking line numbers.

Making static class to hold the `null` AST seems pointless, so this was my workaround.

romani:
regression report will be good to have, as logic is changed

rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/162/

It is showing a ton of differences, so I will need to look into these and see where I went wrong.

rnveach:
http://rveach.no-ip.org/checkstyle/regression/reports/162/openjdk10/index.html#A87
New violations. They are invalid because it is a comment in the middle of a method declaration.
I will fix and add a test case.

rnveach:
> http://rveach.no-ip.org/checkstyle/regression/reports/162/openjdk10/index.html#A87

Fixed and added new test cases. Re-running regression.

romani:
from travis:
`Compilation failure
/home/travis/build/sevntu-checkstyle/sevntu.checkstyle/sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/InputForbidCCommentsInMethodsCheck.java:[84,10] ';' expected
`

rnveach:
New Regression: http://rveach.no-ip.org/checkstyle/regression/reports/163/

rnveach:
@romani I looked through the majority of the violations.
I went through half of the openjdk violations for my first post on regression. I checked through alot of the smaller sets and picked random ones to review.

I don't see anything else so I think we are good. Please review.

Issue #661, opened 2018-05-27
LogicConditionNeedOptimization: ignore instanceof


rnveach:
````
$ cat TestClass.java
public class TestClass {
    void method() {
String s = "";
Object o = null;
if (("Package".equals(s)) && (o instanceof String)) {}
    }
}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
<module name="com.github.sevntu.checkstyle.checks.coding.LogicConditionNeedOptimizationCheck" />
    </module>
</module>

$ java -jar checkstyle-8.9-sevntu-1.29.0-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[ERROR] TestClass.java:5: Condition with && at line 5 position 26 need optimization. All method calls are advised to move to end of logic expression. [LogicConditionNeedOptimization]
Audit done.
Checkstyle ends with 1 errors.
````

`instanceof` should be treated the same as a method call as it is a heavy procedure.

PR #678, opened 2018-06-08
Issue #661: ignore instanceof in LogicConditionNeedOptimizationCheck

rnveach:
Issue #661

Regression to come.

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/LogicConditionNeedOptimizationCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/LogicConditionNeedOptimizationCheck.java
@@ -26,7 +26,7 @@ import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 /**
  * <p>
  * This check prevents the placement of local variables and fields after calling
- * methods in '&amp;&amp;' and '||' conditions.
+ * methods and instanceof in '&amp;&amp;' and '||' conditions.
  * </p>
  * <p>
  * For example: if(getProperty() &amp;&amp; property) ==&gt; if(property &amp;&amp; getProperty()),
@@ -76,8 +76,21 @@ public class LogicConditionNeedOptimizationCheck extends AbstractCheck {
      */
     private static boolean needOptimization(DetailAST logicNode) {
         final DetailAST secondOperand = getSecondOperand(logicNode);
-        return !secondOperand.branchContains(TokenTypes.METHOD_CALL)
-                && logicNode.branchContains(TokenTypes.METHOD_CALL);
+        final boolean firstInstanceOf = logicNode.branchContains(TokenTypes.LITERAL_INSTANCEOF);
+        final boolean secondTypeCast = secondOperand.branchContains(TokenTypes.TYPECAST);
+        final boolean result;
+
+        if (firstInstanceOf && secondTypeCast) {
+            result = false;
+        }
+        else {
+            result = !secondOperand.branchContains(TokenTypes.METHOD_CALL)
+                && !secondOperand.branchContains(TokenTypes.LITERAL_INSTANCEOF)
+                && (firstInstanceOf
+                        || logicNode.branchContains(TokenTypes.METHOD_CALL));
+        }
+
+        return result;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


romani:
@rnveach ,

few more test inputs:
```
if ((o instanceof String)) {} // OK
if ( o==null && (o instanceof String)) {} // OK
if ( (o instanceof String) && o==null ) {} //VIOLATION
if ( o==null && (o instanceof String) && o!=null) {} // VIOLATION
if ( o==null && ((o instanceof String) && o!=null)) {} // VIOLATION
if ( o==null && ((o instanceof String) || o!=null)) {} // OK
if ( o==null && (o!=null && (o instanceof String))) {} // OK
```

rnveach:
> if ( o==null && ((o instanceof String) || o!=null)) {} // OK
if ( o==null && (o!=null && (o instanceof String))) {} // OK

This is wrong. First one should have a violation because `instanceof` is before a simple null check inside the `||` condition.
Second one is correct and shows how statement should be written.

romani:
your corrections are confirmed, you are right.

Lets have a look at regression,

rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/164/

http://rveach.no-ip.org/checkstyle/regression/reports/164/apacheapex/index.html#A5
> if (t instanceof TestControlTuple && ((TestControlTuple)t).data == this.data) {

This is a false positive. `instanceof` is needed first to insure type can be cast.

romani:
Yes, `instanceof` allow to write some special code further in expression, we might need to skip all related casts to same type. This seriously complicate logic as you need to thoroughly analyse whole expression.

rnveach:
The other option is to ignore the violation if the first has `instanceof` and the second has a type-cast.

We could check type being the same as cast, but there could be weird reasons they aren't the same.
The appropriate way would be to verify the variable between the 2 operands are the same with the type-cast check.

I will just add the type-cast check for now. If it is an issue later, we can look into it more.

romani:
> The appropriate way would be to verify the variable between the 2 operands are the same, along with the type-cast check and the variable is in the type-cast.

this is good.
skip all(in expression branch) that is followed instanceof is also ok. 
Any other solution that does not produce false-positives is ok. 
False-negative is much better, as it never fail a build :)

rnveach:
Updated Regression: http://rveach.no-ip.org/checkstyle/regression/reports/166/

rnveach:
Thanks for reviewing. I went to bed as soon as the report was ready.

http://rveach.no-ip.org/checkstyle/regression/reports/166/openjdk10/index.html#A20
> 3138         if ((t instanceof ForkJoinWorkerThread) &&
3139             (p = (wt = (ForkJoinWorkerThread)t).pool) != null &&
3140             (w = wt.workQueue) != null) {

This is a false positive because `wt` is assigned in condition 2 and then used in condition 3.
AST sees conditions as `(cond1 && cond2) && cond3` instead of `cond1 && cond2 && cond3`. Since condition 1 has an `instanceof` and condition 3 doesn't have a typecast, it thinks it can swap the 2 in favor of `cond3 && (cond1 && cond2)`.

This seems like a general issue and not directly related to my changes.

````
$ cat TestClass.java
public class TestClass {
    private int field = 1;

    void method() {
TestClass s = new TestClass(), t = null, u = null;

        if (((t = s).test()) &&
            ((u = t).field == 1) != null &&
            (u.field) != 2) {}
    }

    public boolean test() { return true; }
}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
<module name="com.github.sevntu.checkstyle.checks.coding.LogicConditionNeedOptimizationCheck" />
    </module>
</module>

$ java -jar checkstyle-8.9-sevntu-1.29.0-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[ERROR] TestClass.java:7: Condition with && at line 7 position 29 need optimization. All method calls are advised to move to end of logic expression. [LogicConditionNeedOptimization]
[ERROR] TestClass.java:8: Condition with && at line 8 position 41 need optimization. All method calls are advised to move to end of logic expression. [LogicConditionNeedOptimization]
Audit done.
Checkstyle ends with 2 errors.
````

Re-arranging any of the conditions will cause an NPE.

rnveach:
> Please add it to UT

Done.

I have verified much of openjdk10 as it usually has interesting cases. I also checked the ones you didn't list and I didn't see anything new.

romani:
> Assignments in conditions are bad practice anyways.

yes, could be done separately.

=============

ok to merge.

Issue #659, opened 2018-04-11
return of AvoidNotShortCircuitOperatorsForBoolean reports multi catch case


toli:
seems that another variant of #559 is back. 

seems that if i have a complex multi-line command before the catch with multiple exceptions i get the same incorrect behavior
attaching a pom.xml, java file and the config.xml to run `mvn checkstyle:checkstyle`

/var/tmp $ javac YOUR_FILE.java
compiles via Maven

/var/tmp $ cat YOUR_FILE.java
```
import com.spotify.docker.client.DockerClient;
import com.spotify.docker.client.exceptions.DockerException;
import com.spotify.docker.client.messages.ContainerConfig;
import com.spotify.docker.client.messages.HostConfig;

import java.util.concurrent.ExecutorService;

public class Toli {
  // this will obviously not run at all..
  private void updateDelegateViaDocker(String newVersion) {
    ExecutorService executorService = null;
    executorService.submit(() -> {
      try {
        HostConfig hostConfig = null;
        ContainerConfig containerConfig = ContainerConfig.builder()
            .image("toliImage")
            .attachStderr(true)
            .hostConfig(HostConfig.builder().build()).build();
        DockerClient docker = null;
        docker.startContainer(docker.createContainer(containerConfig).id());

      } catch (InterruptedException | DockerException ex) {
        // boo
      }
    });
  }
}
```
/var/tmp $ cat config.xml
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE module PUBLIC
        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <module name="TreeWalker">
        <module name="AvoidNotShortCircuitOperatorsForBoolean"/>
        <module name="com.github.sevntu.checkstyle.checks.coding.AvoidNotShortCircuitOperatorsForBooleanCheck"/>
    </module>
</module>
```

For Linux users:

```
java -classpath sevntu-checks-1.29.0.jar:checkstyle-8.9-all.jar com.puppycrawl.tools.checkstyle.Main -c config.xml src/main/java/Toli.java
Starting audit...
[ERROR] /private/tmp/checkstyle/src/main/java/Toli.java:22:37: Short-circuit operator should be used instead of '|'. [AvoidNotShortCircuitOperatorsForBoolean]
Audit done.
Checkstyle ends with 1 errors.
```
[...]

PR #679, opened 2018-06-08
Issue #659: fixed AvoidNotShortCircuitOperatorsForBoolean multi-catch

rnveach:
Issue #659 

Issue is not yet approved, but I verified the issue.
Regression to come.


romani:
approved.

regression diff report is required.

Issue #572, opened 2017-05-12
OverridableMethodInConstructorCheck: can't run regression on openjdk9


rnveach:
Config:
````
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>
    <property name="haltOnException" value="false"/>

    <!--
      this config is a copy of
      https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle_sevntu_checks.xml
      with update to global severity=ignore to catch only exceptions and changed to full package names.

      Some other checks are commented out to speed up execution time.
    -->

    <property name="severity" value="ignore"/>

    <!-- All checks are from https://github.com/sevntu-checkstyle/sevntu.checkstyle project -->

	<!-- BeforeExecutionFileFilters is required for sources of java9 -->
    <module name="BeforeExecutionExclusionFileFilter">
        <property name="fileNamePattern" value="module\-info\.java$" />
    </module>

    <module name="TreeWalker">
        <module name="com.github.sevntu.checkstyle.checks.coding.OverridableMethodInConstructor">
            <property name="checkCloneMethod" value="true"/>
            <property name="checkReadObjectMethod" value="true"/>
            <property name="matchMethodsByArgCount" value="true"/>
        </module>
    </module>
</module>
````

Run:
````
Exception in thread "main" java.lang.Error: Error was thrown while processing /home/ricky/opensource/contribution/checkstyle-tester/src/main/java/src/java.base/share/classes/sun/net/www/protocol/jar/JarURLConnection.java
    at com.puppycrawl.tools.checkstyle.Checker.processFiles(Checker.java:297)
    at com.puppycrawl.tools.checkstyle.Checker.process(Checker.java:211)
    at org.apache.maven.plugin.checkstyle.exec.DefaultCheckstyleExecutor.executeCheckstyle(DefaultCheckstyleExecutor.java:252)
    at org.apache.maven.plugin.checkstyle.AbstractCheckstyleReport.executeReport(AbstractCheckstyleReport.java:473)
    at org.apache.maven.plugin.checkstyle.CheckstyleReport.executeReport(CheckstyleReport.java:154)
    at org.apache.maven.reporting.AbstractMavenReport.generate(AbstractMavenReport.java:255)
[...]

PR #681, opened 2018-06-08
Issue #572: fixed infinite loop in OverridableMethodInConstructorCheck

rnveach:
Issue #572 

Added fix to prevent infinite loop.

Will provide limited difference regression and no exception report.


rnveach:
No Exception Test: http://rveach.no-ip.org/checkstyle/regression/reports/168/

Check produced no exceptions. Next is limited regression run.

rnveach:
Limited Regression Report: http://rveach.no-ip.org/checkstyle/regression/reports/169/

No differences.

Issue #683, opened 2018-06-08
LogicConditionNeedOptimization: false positive when many conditions are wrapped in parenthesis


rnveach:
````
$ cat TestClass.java
import java.nio.*;

public class NativeOrder {
    public static void main(String[] args) throws Exception {
        ByteOrder bo = ByteOrder.nativeOrder();
        System.err.println(bo);
        String arch = System.getProperty("os.arch");
        if (((arch.equals("sparc") && (bo != ByteOrder.BIG_ENDIAN))) ||
            ((arch.equals("i386") && (bo != ByteOrder.LITTLE_ENDIAN)))) {
            throw new Exception("Wrong byte order");
        }
    }
}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
<module name="com.github.sevntu.checkstyle.checks.coding.LogicConditionNeedOptimizationCheck">
</module>
    </module>
</module>

$ java -jar checkstyle-8.9-sevntu-1.29.0-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[ERROR] TestClass.java:8: Condition with && at line 8 position 35 need optimization. All method calls are advised to move to end of logic expression. [LogicConditionNeedOptimization]
[ERROR] TestClass.java:8: Condition with || at line 8 position 69 need optimization. All method calls are advised to move to end of logic expression. [LogicConditionNeedOptimization]
[ERROR] TestClass.java:9: Condition with && at line 9 position 34 need optimization. All method calls are advised to move to end of logic expression. [LogicConditionNeedOptimization]
Audit done.
Checkstyle ends with 3 errors.
````

Violations for 8:35 and 9:34 are valid.
The violation at 8:69 is not. The `||` between the 2 big blocks both have method calls, so there is no better ordering switching them around, and the violation will stay even if they are switched.
Once the other 2 violations are fixed, the other one disappears.

I couldn't create an example that only involved 1 violation.

rnveach:
The issue is LPAREN and RPAREN aren't nested in each other in this type of situation.

````
    |      |  |  `--LOR -> || [8:69]
    |      |  |      |--LPAREN -> ( [8:12]
[...]

PR #684, opened 2018-06-09
Issue #683: fixed parenthesis handling in LogicConditionNeedOptimization

rnveach:
Issue #683

Regression: http://rveach.no-ip.org/checkstyle/regression/reports/171/

All are false positives being removed.


romani:
trend is good, less false-positives the better.

but his looks like should stay as violation.
http://rveach.no-ip.org/checkstyle/regression/reports/171/openjdk10/index.html#A7

but we can merge this RP , and improve in next PR.
One more time:  false-negative is better than false-positive.

rnveach:
> but his looks like should stay as violation.

This violation is on the `||` and it is a false positive and is the exact same example that was given in the issue and placed in the UT.

> 38          if (((arch.equals("i386") && (bo != ByteOrder.LITTLE_ENDIAN))) ||
39              ((arch.equals("amd64") && (bo != ByteOrder.LITTLE_ENDIAN))) ||

code is `(a && b) || (c && d)....`. `b` and `d` are simple checks with no methods, so they should be swapped, but there is no difference between `a && b` and `c && d`. Both house a simple check and a method call. There is no better ordering between `(c && d) || (a && b)` or the original.

Issue #650, opened 2018-03-04
RedundantReturn: false positive in anonymous class


rnveach:
````
$ cat TestClass.java
public class TestClass {
    private static void writePropertiesFile(File inputFile, File outputFile)
            throws CheckstyleException {
        PrintWriter writer = null;
        try {
            writer = new PrintWriter(outputFile, StandardCharsets.UTF_8.name());

            final DetailAST top = JavaParser.parseFile(inputFile, JavaParser.Options.WITH_COMMENTS);
            final DetailAST objBlock = getClassBody(top);
            if (objBlock != null) {
                final PrintWriter writerEx = writer;

                iteratePublicStaticIntFields(objBlock, new Consumer<String>() {
                    @Override
                    public boolean accept(String s) {
                        writerEx.println(s);
                        return true; // violation
                    }
                });
            }
        }
        catch (IOException ex) {
            throw new CheckstyleException("Failed to write javadoc properties of '" + inputFile
                + "' to '" + outputFile + "'", ex);
        }
        finally {
            CommonUtils.close(writer);
        }
    }
}

$ cat TestConfig.xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">

<module name="Checker">
    <property name="charset" value="UTF-8"/>

    <module name="TreeWalker">
<module name="com.github.sevntu.checkstyle.checks.coding.RedundantReturnCheck"/>
    </module>
</module>

$ java -jar checkstyle-8.8-sevntu-1.28.0-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[ERROR] TestClass.java:17: Redundant return. [RedundantReturn]
Audit done.
Checkstyle ends with 1 errors.
````

The line in question is not redundant and would result in non-compilable code if it was removed or changed to a single `return`.
This issue was found backporting `JavadocPropertiesGenerator` to Java 6 for checkstyle-backport-jre6 .

http://sevntu-checkstyle.github.io/sevntu.checkstyle/apidocs/com/github/sevntu/checkstyle/checks/coding/RedundantReturnCheck.html , redundant `return` is only `return;` all other forms of return is not redundant.

rnveach:
It seems it doesn't realize the separation of the 2 methods by the anonymous class.
[...]

PR #686, opened 2018-06-09
Issue #650: skip over object blocks in RedundantReturnCheck

rnveach:
Issue #650

Regression to come.

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RedundantReturnCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RedundantReturnCheck.java
@@ -327,17 +327,17 @@ public class RedundantReturnCheck extends AbstractCheck {
     private static DetailAST findRedundantReturnInCatch(DetailAST lastStatementInCatchBlockAst) {
         DetailAST redundantReturnAst = null;
         DetailAST currentNodeAst = lastStatementInCatchBlockAst;
-        DetailAST toVisitAst = currentNodeAst;
         DetailAST returnAst = null;
-        while (toVisitAst != null) {
-            toVisitAst = Utils.getNextSubTreeNode(toVisitAst, currentNodeAst);
+        DetailAST toVisitAst = Utils.getNextSubTreeNode(currentNodeAst, currentNodeAst);
 
-            if (toVisitAst != null
-                    && (toVisitAst.getParent().getParent().getNextSibling() == null
-                        || toVisitAst.getParent().getParent().getNextSibling().getType()
-                            == TokenTypes.RCURLY)
-                    && toVisitAst.getType() == TokenTypes.LITERAL_RETURN
-                    && toVisitAst.getParent().getNextSibling() == null) {
+        while (toVisitAst != null) {
+            if (toVisitAst.getType() == TokenTypes.OBJBLOCK) {
+                while (toVisitAst.getNextSibling() == null) {
+                    toVisitAst = toVisitAst.getParent();
+                }
+                toVisitAst = toVisitAst.getNextSibling();
+            }
+            else if (isFinalReturn(toVisitAst)) {
                 returnAst = toVisitAst;
 
                 while (toVisitAst != null
@@ -351,10 +351,25 @@ public class RedundantReturnCheck extends AbstractCheck {
 
                 toVisitAst = returnAst;
             }
+
+            toVisitAst = Utils.getNextSubTreeNode(toVisitAst, currentNodeAst);
         }
 
         currentNodeAst = Utils.getNextSubTreeNode(currentNodeAst, lastStatementInCatchBlockAst);
         return redundantReturnAst;
+    }
+
+    /**
+     * Checks if the {@code ast} is the final return statement.
+     * @param ast the AST to examine.
+     * @return {@code true} if the {@code ast} is the final return statement.
+     */
+    private static boolean isFinalReturn(DetailAST ast) {
+        return (ast.getParent().getParent().getNextSibling() == null
+                    || ast.getParent().getParent().getNextSibling().getType()
+                    == TokenTypes.RCURLY)
+                && ast.getType() == TokenTypes.LITERAL_RETURN
+                && ast.getParent().getNextSibling() == null;
     }
 
 }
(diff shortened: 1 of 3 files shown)


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/172/

http://rveach.no-ip.org/checkstyle/regression/reports/172/Hbase/index.html#A1
http://rveach.no-ip.org/checkstyle/regression/reports/172/elasticsearch/index.html#A2
This looks like this violation should have stayed. I will have to look at it closer.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RedundantReturnCheck.java
```diff
@@ -328,15 +328,25 @@ private static DetailAST findRedundantReturnInCatch(DetailAST lastStatementInCat
...
+        while (true) {
             toVisitAst = Utils.getNextSubTreeNode(toVisitAst, currentNodeAst);
 
-            if (toVisitAst != null
-                    && (toVisitAst.getParent().getParent().getNextSibling() == null
-                        || toVisitAst.getParent().getParent().getNextSibling().getType()
-                            == TokenTypes.RCURLY)
-                    && toVisitAst.getType() == TokenTypes.LITERAL_RETURN
-                    && toVisitAst.getParent().getNextSibling() == null) {
+            if (toVisitAst == null) {
```
romani: why not simply to check children on `return`.
```
$ cat Test.java 
public class Test {
    boolean foo() {
        return true;
    }
}

$ java -jar checkstyle-8.10.1-all.jar -t Test.java  | grep "\[3:" | sed -E "s/ +\| +\|//"
--LITERAL_RETURN -> return [3:8]
   |--EXPR -> EXPR [3:15]
   |   `--LITERAL_TRUE -> true [3:15]
   `--SEMI -> ; [3:19]

```

rnveach:
New Regression: http://rveach.no-ip.org/checkstyle/regression/reports/173/

Reports look good to me now.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RedundantReturnCheck.java
rnveach: > why not simply to check children on return

While it would give us the same results, to me it seems it would just mask the issue. We should only be scanning the current method we are in. Once we enter an anonymous class, or possible a lambda, we are scanning outside the scope of what we are checking. My changes have the benefit that we will skip those tokens and their children completely, while verify `return` is by itself would still require us to parse the entire anonymous class.

rnveach: Shouldn't this check have lambda as a token?

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/InputRedundantReturnCheckNestedMethods.java
```diff
@@ -0,0 +1,41 @@
+package com.github.sevntu.checkstyle.checks.coding;
+
+public class InputRedundantReturnCheckNestedMethods {
+    public void method() {
+        try {
+            if (true) {
+                new Nested() {
+                    @Override
+                    public boolean accept(String s) {
+                        return true; // violation
```
romani: why violation is here ?

redundant return is only "return;". Am i miss smth ?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RedundantReturnCheck.java
rnveach: > It force code to have break in code it is the same as multiple exit points from method, or multiple breaks in loops. 

Each loop has 1 break otherwise `SingleBreakOrContinueCheck` would have violated it.

> l dislike and mistrust for while (true)

nested loop `while true` can be changed to:

````
                do {
....
                } while (toVisitAst != null);
````

But we have nothing to make the condition false for code coverage.

1st loop can be changed to:

````
        DetailAST toVisitAst = Utils.getNextSubTreeNode(currentNodeAst, currentNodeAst);
        while (toVisitAst != null) {
....
            toVisitAst = Utils.getNextSubTreeNode(toVisitAst, currentNodeAst);
        }
````

And cover all code coverage.
Do you prefer these more?

Comment (romani) [reply sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RedundantReturnCheck.java]: > Do you prefer these more?

yes, please remove `while (true)`.


>But we have nothing to make the condition false for code coverage

Following code passed UT and coverage on my side:
```
                while (toVisitAst.getNextSibling() == null) {
                    toVisitAst = toVisitAst.getParent();
                }
                toVisitAst = toVisitAst.getNextSibling();
```

romani:
report looks good.
Please resolve `while (true)` and others.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RedundantReturnCheck.java
rnveach: Done.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/InputRedundantReturnCheckNestedMethods.java
rnveach: Done.

romani:
Does it make sense to recheck child token under return ? If there are some not semi - it is not redundant.

rnveach:
The following code compiles for me, so lambdas can have empty return.
````
class Test2 {
	public interface Something {
		void action();
	}

	void methodL() {
		Something l = () -> {
			System.out.println("Do nothing!");
			return;
		};
	}
}
````

rnveach:
@romani Actually lambda doesn't have type, so when we implement lambda we will handle empty return.

romani:
> but I am thinking we will hit it for lambdas, so once we add that token, we are back to not being able to cover it.

such request moved to #700 

--------------

@rnveach , as we did a refactoring, please regenerate regression, if it is the same - ok to merge

rnveach:
New Report: http://rveach.no-ip.org/checkstyle/regression/reports/178/

Nothing has changed since last report.

PR #688, opened 2018-06-09
Issue #687: fixed NPE with empty methods in RedundantReturnCheck

rnveach:
Issue #687 

Regression to come.


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/174/

Removed alot of exceptions which were hiding violations.
Travis failure is because issue isn't approved.

PR #689, opened 2018-06-09
config: added skip option to travis

rnveach:
Added skip capability to sevntu.

@romani I don't think there are any ones that should be forced to always run, but let me know if you think some should always be forced.


rnveach:
wercker will fail until YML is added.

romani:
one travis , pass - ok to merge.

Issue #644, opened 2018-01-22
Add no-exceptions testing lauches to CI


romani:
when new Check is proposed in PR, after code-review almost forgot to request regression testing.
At least default execution of check should be done automatically on most our testing projects, as we do this at:
https://github.com/checkstyle/checkstyle/blob/master/wercker.yml

the same CIs could be used.

rnveach:
>   groovy ./launch.groovy --listOfProjects projects-for-wercker.properties \
      --config checks-nonjavadoc-error.xml --checkstyleVersion ${SEVNTU_POM_VERSION}

Groovy scripts don't support `sevntuVersion`.

https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/pom.xml#L16
Default maven configuration uses release version, not snapshot version.
https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/pom.xml#L15
Default maven configuration uses snapshot version for checkstyle.

We need to be able to override both values.

romani:
Fix is merged

PR #693, opened 2018-06-09
Issue #644: added no exception testing to wercker

rnveach:
Issue #644

### DIFF
--- a/sevntu-checks/.ci/wercker.sh
+++ b/sevntu-checks/.ci/wercker.sh
@@ -1,0 +1,188 @@
+#!/bin/bash
+# Attention, there is no "-x" to avoid problems on Wercker
+set -e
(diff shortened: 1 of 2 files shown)


rnveach:
> Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Unable to instantiate 'com.github.sevntu.checkstyle.checks.design.StaticMethodCandidate' class, it is also not possible to instantiate it as null. Please recheck that class name is specified as canonical name or read how to configure short name usage http://checkstyle.sourceforge.net/config.html#Packages. Please also recheck that provided ClassLoader to Checker is configured correctly.

Configuration needs more fixing.

rnveach:
@romani CI passed.

PR #694, opened 2018-06-09
minor: removed unnecessary file

rnveach:


romani:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/sevntu-checks.xml#L7

Is it the same file ?

rnveach:
I forgot there is a different command to execute sevntu on itself. I thought this file wasn't used since full verify didn't complain.

romani:
We might update project to run most of validation in verify phase.
It better to know only ` mvn verify` to run all validations from pom.

PR #697, opened 2018-06-09
Issue #525: removed inner assignment suppression and fixed violations

rnveach:
Issue #525


rnveach:
I will run regression just to make sure there is no issues.

rnveach:
http://rveach.no-ip.org/checkstyle/regression/reports/175/

No differences.

PR #699, opened 2018-06-10
Issue #525: removed some UnnecessaryParenthesesExtendedCheck violations

rnveach:
Issue #525


romani:
as it is not simple extraction of code to function , please run regression.

if no regression  - ok to merge.

rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/176/

No differences.

PR #474, opened 2016-10-20
Issue #464: Check for trailing commas on enums

kariem:
This Check assures that the last enum constant definition in a multi-line enum ends in a comma:

    enum E1 {
        ONE,
        TWO,
        THREE,
    }

Supersedes #465

Usage examples are in the Javadoc of `EnumTrailingCommaCheck`. This check follows the concept of [ArrayTrailingComma](http://checkstyle.sourceforge.net/config_coding.html#ArrayTrailingComma), hence the similar name.

It also checks that the "trailing" comma after the last enum constant definition is not followed by a semicolon, because this would miss the point of the check: clean diff and easy manipulation.

Issue #464, opened 2016-07-22
New check: Trailing comma in enum similar to array initializer


kariem:
_copied from checkstyle/checkstyle#3161_

Similar to [ArrayTrailingComma](http://checkstyle.sourceforge.net/config_coding.html#ArrayTrailingComma) there should be a check for trailing commas in enum constants. A lot of enums are used to map to values from somewhere else and it might be necessary to adapt them often.

The check verifies that the last element of a multi-line enum constant definition has a comma:

Check ok:
```java
enum MovieTypes {
  GOOD,
  BAD,
  UGLY,
  ;
}
```

Check failed
```java
enum ErrorMessages {
  USER_DOES_NOT_EXIST,
  INVALID_USER_PASS,
  EVERYTHING_ELSE;
}
```

I stumbled across this while [writing](http://kariem.github.io/2016/05/04/trailing-comma-java/) how to check for trailing commas in arrays and enums based on an article discussion JavaScript.

romani:
please review recent issues to see what is required to discuss (names, config) before implementation to avoid extra coding/refactoring/.... .

romani:
Referenced PR was not finished (but functionally it works, but still require minor changes to be completed) and abandoned .... . Everyone is welcome to continue.

romani:
https://checkstyle.org/config_coding.html#UnnecessarySemicolonInEnumeration is about `;` .... so Check for trailing comma `,` is still in demand.

DmytroBarabash:
Hey @kariem @yaziza 
I think it is a great idea to implement this Check (I even created a similar issue, but @rnveach suggested writing here first)
Maybe we need one more small step to finish this job eventually. I'd like to take part in it if possible.
What blocks the final implementation/merge now, and how could I help?

kariem:
@DmytroBarabash I have created #465, #474, #709, #722 to address this issue and with every change the maintainers found new things that needed to be addressed before merging anything. Effectively, this was not much work on the implementation side. However, the requests in the PRs increased the required work by at least a magnitude. Instead of just changing wording here or there to better fit with what the maintainers expected, there were just more and more changes after addressing an issue. I did not see an end to the additional requested changes and it did not seem as if the work put into the changes were appreciated at all. The last comment even suggested to just [rename](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/722#discussion_r228189882) everything, more than 2 years after the first commit.

[...]

PR #709, opened 2018-07-26
Issue #464: Check for trailing commas on enums

kariem:
Addresses #464, supersedes #465 and #474.

This Check assures that the last enum constant definition in a multi-line enum ends in a comma:

```
enum E1 {
    ONE,
    TWO,
    THREE,
}
```

Usage examples are in the Javadoc of `EnumTrailingCommaCheck`. This check follows the concept of [ArrayTrailingComma](http://checkstyle.sourceforge.net/config_coding.html#ArrayTrailingComma), hence the similar name.

It also checks that the "trailing" comma after the last enum constant definition is not followed by a semicolon, because this would miss the point of the check: clean diff and easy manipulation.


rnveach:
The following are the CI failures.

> All sevntu checks should be used in contribution

Check must be added to https://raw.githubusercontent.com/checkstyle/contribution/master/checkstyle-tester/checks-sevntu-error.xml .

> checkstyle-regression
> Unable to instantiate 'com.github.sevntu.checkstyle.checks.coding.EnumTrailingCommaAndSemicolonCheck' class
> wercker
> org.eclipse.aether.resolution.ArtifactResolutionException: Failure to find com.github.sevntu-checkstyle:sevntu-checkstyle-maven-plugin:jar:1.30.0

I will have to see what is wrong.

Checkstyle regression will require changes in main checkstyle project based on violations from this check.
@romani Confirm you are ok with these type of changes in main project.

romani:
@kariem ,


>Checkstyle regression will require changes in main checkstyle project based on violations from this check.

Please provide update to our code in separate commit, it better to make it first commit in this PR, as it unlikely to change. Such commit culd be merged even before main fix to ease merge of new Check.

romani:
Hmm, a lot time passes from original discussion, originally reported at https://github.com/checkstyle/checkstyle/issues/3161. We are not in moratorium any more, so if @kariem ok it is better to contribute to main project directly. But we can finish contribution to sevntu first, whatever is easier.

@rnveach , please review main issue and approve it.

rnveach:
> checkstyle-regression
Unable to instantiate 'com.github.sevntu.checkstyle.checks.coding.EnumTrailingCommaAndSemicolonCheck' class

This was fixed. CI now shows results of regresion. Main checkstyle project has 27 violations from this new check.
See https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/jobs/408476345#L1063

romani:
Wercker failure is strange, I cleaned cache and relaunched build.

rnveach:
wercker passed, so CI validation is working correctly now.

kariem:
@romani Sorry, I do not understand what needs to be done. If I have to change the checkstyle code, I will have to do this in another PR on [checkstyle/checkstyle](https://github.com/checkstyle/checkstyle), right?

What's your recommendation? I think you guys can better assess what is the best way forward.

rnveach:
> I will have to do this in another PR on checkstyle/checkstyle

Correct. Sevntu's checks are eventually applied to checkstyle main project.

> What's your recommendation?

It's best to make this CI green to merge, so we should do PR in main repo.

romani:
I suggest to finish merge to sevntu, as a lot already done and we are close to merge state.
Moving to main repo could cost extra changes for quality, a quality level in main repo significantly higher.

It is better to merge, this PR sooner.

rnveach:
@romani Then it is your call on what to do in this PR. CI will never be green without changes to main repo and external file.

romani:
Let's merge new Check to sevntu first.


Failure in checkstyle-regression 
https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/jobs/408476345#L1028 should be fixed in code of main repo. @kariem , please send PR to main repo with fixes for missed comma. We we quickly merge it .

> Failure "All sevntu checks should be used in contribution"

Please send PR to https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/checks-sevntu-error.xml
We will merge it right after release of your new Check. It is ok to keep this CI point to be failed, it is not a blocked for merge.

kariem:
> please send PR to main repo with fixes for missed comma. We we quickly merge it.

I have created a pull request in checkstyle/checkstyle (referenced above). However, there seem to be conflicts with other checks. Is it necessary to enable *all* checks? I think it's quite obvious that some checks do not play well with others.

> Please send PR to https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/checks-sevntu-error.xml

Does that mean that I should add `EnumTrailingCommaAndSemicolonCheck` to this configuration?

romani:
>I think it's quite obvious that some checks do not play well with others.

I didn't predicted a conflict , so it is not obvious. Need to be warned in documentation as minimum.

>Does that mean that I should add EnumTrailingCommaAndSemicolonCheck to this configuration?

Yes, please do, sorry for confusion.

kariem:
>> I think it's quite obvious that some checks do not play well with others.

> I didn't predicted a conflict , so it is not obvious.

I was talking about the general implication of requirements for code formatting. It is obvious to me that some checks have conflicting requirements. In this special case, I did not know that there were conflicts until I have seen them. However, I assumed that you were aware of checks that are already in place in CI.

> Need to be warned in documentation as minimum.

Done: b54be53d6f1b094d9570b225abae3d137d8df6a1. Is this sufficient for you?



I will add to the [checkstyle/contribution/checkstyle-tester/checks-sevntu-error.xml](https://github.com/checkstyle/contribution/blob/789055b817de65290d31c125f5b7cb256679197f/checkstyle-tester/checks-sevntu-error.xml) as soon as we have [agreed on a solution](https://github.com/checkstyle/checkstyle/pull/6054#issuecomment-409509815) to the conflicting checks. I hope this also makes sense to you.

kariem:
>> I think it's quite obvious that some checks do not play well with others.

> I didn't predicted a conflict , so it is not obvious.

I was talking about the general implication of requirements for code formatting. It is obvious to me that some checks have conflicting requirements. In this special case, I did not know that there were conflicts until I have seen them. However, I assumed that you were aware of checks that are already in place in CI.

> Need to be warned in documentation as minimum.

Done: 5917f4f12d410ac222121e6d6a818d30706209db. Is this sufficient for you?

kariem:
I will add to the [checkstyle/contribution/checkstyle-tester/checks-sevntu-error.xml](https://github.com/checkstyle/contribution/blob/789055b817de65290d31c125f5b7cb256679197f/checkstyle-tester/checks-sevntu-error.xml) as soon as we have [agreed on a solution](https://github.com/checkstyle/checkstyle/pull/6054#issuecomment-409509815) to the conflicting checks. I hope this also makes sense to you.

romani:
>Is this sufficient for you?

Good. Please squash commits in one.

>conflicting checks. I hope this also makes sense to you.

Please add it with ignore severity. Looks like Check is too conflicting now.

rnveach:
@kariem To disable regression for your new check, please add it to the ignore list at:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java#L37

rnveach:
> I will add to the checkstyle/contribution/checkstyle-tester/checks-sevntu-error.xml as soon as we have agreed on a solution to the conflicting checks. 

The check can still be added, as is, to this configuration. The configuration already sets severity to ignore for all checks. See https://github.com/checkstyle/contribution/blob/789055b817de65290d31c125f5b7cb256679197f/checkstyle-tester/checks-sevntu-error.xml#L15 .

kariem:
I have now disabled the check for regression as [instructed by rnveach](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/709#issuecomment-409712433) and squashed all commits into one.

I have created a separate issue and PR on [checkstyle/contribution](https://github.com/checkstyle/contribution) for inclusion of EnumTrailingCommaAndSemicolonCheck.

@rnveach @romani Please let me know, if there is anything else I can do now.

romani:
I need some time to review, I am afk till Tuesday .

Side note is that logic should be "similar" to http://checkstyle.sourceforge.net/config_coding.html#ArrayTrailingComma

Looks like trailing `;` is optional if nothing else except elements defined.

kariem:
@romani You need anything from me?

Should we change the name back to `EnumTrailingComma` to make the relation to [ArrayTrailingComma](http://checkstyle.sourceforge.net/config_coding.html#ArrayTrailingComma) clear?

romani:
@kariem , sorry for long delay in reply.

Please change commit message to "Issue #464: Check for trailing commas on enums." , no need to reference old PRs, you can reference it in issue to keep track/connected.

rnveach:
> Travis failure is reproduced on my local on your branch (not in master):
> `<module name="com.github.sevntu.checkstyle.checks.coding.EnumTrailingCommaAndSemicolonCheck"></module>`

@romani This is because this is the full regression test of new modules in Sevntu run on Checkstyle.
Specifically this happens in https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java#L104 . We pull in checkstyle's repo and modify `checkstyle_sevntu_checks.xml` to include all new modules in sevntu PR that are not in checkstyle. This is how we know if a new check causes any violations on sevntu once integrated.

Because of formatting restrictions places on the XML files now, `CheckstyleRegressionTest` will have to be modified to split the module name up since we use the full package name.

Travis run shows this is the only violation, so it is safe to ignore the failure for now. Once this module is merged, sevntu is released, and we add the new module to checkstyle the issue will go away for now.

romani:
@rnveach , it took me some time to actually understand a reason.
Can we update `setupFiles`([in build](https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/.ci/travis.sh#L80), [in code](https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java#L104)) to put module name in multiline:
```
<module name=
"com.github.sevntu.checkstyle.checks.coding.EnumTrailingCommaAndSemicolonCheck"
></module>
```

so we will be safe in future, no matter what size of name we have.

rnveach:
> Can we update to put module name in multiline

Yes we can.

> will be safe in future, no matter what size of name 

We may have an issue again if package name is made longer, or if Check name is long.
`"com.github.sevntu.checkstyle.checks.coding.` and `"` is 45 characters (50 with our longest package name currently). 
If check name is ever longer then 50-55 characters, it will still cause an issue again.

romani:
> If check name is ever longer then 50-55 characters, it will still cause an issue again

it will be a problem in `checkstyle`'s CI  after new Check introduced. I hope we will keep reasonable on names.
Are you ok to create separate issue on this(update in RegressionTest) ?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaAndSemicolonCheck.java
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * <p>Please note that using this check together with {@code NoWhitespaceBefore} or
+ * {@code SeparatorWrap} may create conflicts with enums that contain a body:
+ * {@code EnumTrailingComma} enforces the semicolon on a separate line while
+ * {@code NoWhiteSpaceBefore} does not allow the semicolon to be preceded with whitespace and
+ * {@code SeparatorWrap} expects the semicolon to be on the same line as the last enum constant.
+ *
+ * @author <a href="kariem.hussein@gmail.com">Kariem Hussein</a>
+ */
+public class EnumTrailingCommaAndSemicolonCheck extends AbstractCheck {
```
romani: @kariem , yes you are right we need to rename Check to `EnumTrailingComma`  as you already did in javadoc example :) .
this Check is focused on trailing comma to be the last symbol inline. Occasional semicolon is just a result of rule "comma should be the last symbol in line".

kariem:
> Please change commit message to "Issue #464: Check for trailing commas on enums." , no need to reference old PRs, you can reference it in issue to keep track/connected.

Done: c874700712c620a74f195a0a3fc4aaaffd3efa3e

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaAndSemicolonCheck.java
kariem: Separate commit with rename OK, or should I amend the current commit? I will rename all occurrences.

kariem:
@romani 

> @kariem , yes you are right we need to rename Check to `EnumTrailingComma` as you already did in javadoc example :) .
> this Check is focused on trailing comma to be the last symbol inline. Occasional semicolon is just a result of rule "comma should be the last symbol in line".

Done: fbe285723a550d7b9f4f5487530993d7903ebfec

I have also adapted the corresponding commit in checkstyle/contribution#333

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaAndSemicolonCheck.java
romani: In PR only one commit.
All should be squashed in one single commit.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
```diff
@@ -0,0 +1,159 @@
...
+        final DetailAST enumConstLeft = enumConstBlock.findFirstToken(TokenTypes.LCURLY);
+        final DetailAST enumConstRight = enumConstBlock.findFirstToken(TokenTypes.RCURLY);
+
+        // Only check, if block is multi-line and there are more than one enum constants
+        if (enumConstLeft.getLineNo() != enumConstRight.getLineNo()
+                && enumConstBlock.getChildCount(TokenTypes.ENUM_CONSTANT_DEF) > 1) {
+            final DetailAST constant = enumConstBlock.findFirstToken(TokenTypes.ENUM_CONSTANT_DEF);
+            final DetailAST lastComma = getLastComma(constant);
+
+            final DetailAST next = lastComma.getNextSibling();
```
romani: next => nextAst
type => nextAstType

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
```diff
@@ -0,0 +1,159 @@
...
+ * {@code SeparatorWrap} expects the semicolon to be on the same line as the last enum constant.
+ *
+ * @author <a href="kariem.hussein@gmail.com">Kariem Hussein</a>
+ */
+public class EnumTrailingCommaCheck extends AbstractCheck {
+
+    /** Key for warning message text in "messages.properties" file. */
+    public static final String MSG_KEY = "enum.trailing.comma";
+    /** Key for warning message text in "messages.properties" file. */
+    public static final String MSG_KEY_SEMI = "enum.trailing.comma.semi";
```
romani: it is not clear why you need two different messages.
rule is simple ... smth right after `,` -> violation

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
```diff
@@ -0,0 +1,159 @@
...
+            if (type == TokenTypes.SEMI) {
+                if (next.getLineNo() == lastComma.getLineNo()) {
+                    // semi on the same line as last comma
+                    log(next.getLineNo(), MSG_KEY_SEMI);
+                }
+
+            }
+            else if (type == TokenTypes.ENUM_CONSTANT_DEF) {
+                log(next.getLineNo(), MSG_KEY);
+            }
```
romani: code should be like:
```
if ( isSmthAfterComm(lastComma, nextAst) ){
   log(next.getLineNo(), MSG_KEY);
}
```

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    /**
+     * Get the last comma in a series of siblings.
+     *
+     * @param start the first sibling
+     * @return the AST containing the last comma
+     */
+    private DetailAST getLastComma(DetailAST start) {
```
romani: please do me a favor ....
make it as static method like `getLast(start, TokenType.Comma)` it might be moved eventually to some Utils class.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
```diff
@@ -0,0 +1,159 @@
...
+
+import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
+import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.TokenTypes;
+
+/**
+ * Checks if enum constant contains an optional trailing comma.
+ *
+ * <p>Rationale: Putting this comma in makes is easier to change the order of the elements or add
+ * new elements at the end of the list.
```
romani: add new sentence "Similar to https://checkstyle.org/config_coding.html#ArrayTrailingComma"

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: We need to distinguish between

``` java
    enum E1 {
        ONE,
        TWO,
        THREE,
    }
```

and

``` java
    enum E4 {
        ONE,
        TWO,
        THREE,; // violation
    }
```

In addition, there is not much that can come after a comma that follows an enum constant definition: another definition, or a semi colon.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: I don't think this should be part of this PR.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: Done: 47a0f0b7be79714fd71c7dd5403ace47aa9fd749

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: Amended Javadoc in 9d9f8516e98ee11f804ea2e7eb512bf1258b3cb3

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: Done: 5681b033c75b407ed7f3f8b183b49a69f4c6c936

kariem:
@rnveach @romani Please review again. I've added separate commits to let you see changes more clearly. Will squash after your ok.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
romani: make it static in this PR, and moving to another Util file is another PR.

romani:
> I've added separate commits to let you see changes more clearly.

please squash now, I always review all from scratch, as after a fix I see more issue that were not obvious on previous review.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
romani: >Enum constant should contain trailing comma.
 vs
Enum constant should contain trailing comma without semi-colon afterwards.

"Enum constant should contain trailing comma"  can we just state that it should be the last symbol in line of code ?
@rnveach , what do you think ?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: Done: e06d5c8029a93c71f12fc0ff366a92eb683c7de2

kariem:
> please squash now, I always review all from scratch, as after a fix I see more issue that were not obvious on previous review.

Done: 4fc3fcb94388c19168504ec388db67976b73e40d

Please note that you can use the [Files tab](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/709/files) for this without the need for squashing everything before a pull request merge. See also the instructions in the official GitHub documentation [here](https://help.github.com/articles/reviewing-proposed-changes-in-a-pull-request/#starting-a-review), step 3: "On the pull request, click **Files changed**."

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: Going over this message again, after more than two years (af7482f459b006813717399616212aaaa0752a3d), I think the message is not correct.

> Enum constant should contain trailing comma.

Better: "Enum constant should _be followed_ by a trailing comma."

Then, the second, more detailed message makes more sense too:

"Enum constant should _be followed_ by a trailing comma without semi-colon afterwards."

What do you think?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
```diff
@@ -0,0 +1,160 @@
...
+            final DetailAST constant = enumConstBlock.findFirstToken(TokenTypes.ENUM_CONSTANT_DEF);
+            final DetailAST lastComma = getLastComma(constant);
+
+            final DetailAST nextAst = lastComma.getNextSibling();
+            final int nextAstType = nextAst.getType();
+
+            if (nextAstType == TokenTypes.SEMI) {
+                if (nextAst.getLineNo() == lastComma.getLineNo()) {
+                    // semi on the same line as last comma
+                    log(nextAst.getLineNo(), MSG_KEY_SEMI);
```
rnveach: Log the entire AST, not just the line.
Without the full AST, xpath suppressions won't work.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
```diff
@@ -0,0 +1,160 @@
...
+                }
+
+            }
+            else if (nextAstType == TokenTypes.ENUM_CONSTANT_DEF) {
+                log(nextAst.getLineNo(), MSG_KEY);
```
rnveach: same here.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
rnveach: @romani 
> "Enum constant should contain trailing comma" can we just state that it should be the last symbol in line of code ?

Wouldn't the `/` of a comment be a symbol? Check isn't comment aware, so it wouldn't throw a violation if a comment appeared after the comma.

rnveach: @romani 

> "Enum constant should contain trailing comma" can we just state that it should be the last symbol in line of code ?

Wouldn't the / of a comment be a symbol? Check isn't comment aware, so it wouldn't throw a violation if a comment appeared after the comma.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: Done: 120d4aa2ef75b1e54fe00f375353c2ba95f250f6

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: Done: 120d4aa2ef75b1e54fe00f375353c2ba95f250f6

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: In my last comment, I have suggested improved error messages. What about these?

romani:
> Please note that you can use the

yes, I know, but in my experience demanding it to be squashed by user works better. Looks like mobile version is upgraded to show all combined also. I do not remember what else forced me to demand this step from contributors.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
romani: @kariem , @rnveach 
What about `Enum constant should be followed by a trailing comma only in line` ?

rnveach: The correct wording is `Enum constant should be followed by only a trailing comma in the line`.

@kariem I think @romani is trying to condense the 2 messages into 1.

kariem: @rnveach @romani 

One message is definitely easier to implement. The thinking behind the current implementation was that we show the user the simplest message (`Enum constant should be followed by a trailing comma.`), and, only if necessary, tell them, that a semicolon on the same line is not allowed (`Enum constant should be followed by a trailing comma without semi-colon afterwards`).

Anyways, after your suggestions, how about the following:

`Enum constant should be followed by a trailing comma in the line.`

This the main point of the rule we are trying to enforce. I also think this message is very clear. In addition, if someone puts a semicolon right after the comma, we can show another message for clarity:

`Enum constant should be followed by a trailing comma in the line and no additional semicolon.`

The difference is that this is good

    enum E {
        ONE,
        TWO, // some comment
        THREE,
    }

while this is bad

    enum E {
        ONE,
        TWO,
        THREE,;
    }

The first allows for better diffs when adding or removing an enum constant, while the latter does not. This is why we should not write "by _only_ a trailing comma". It's not correct, as a comment is also fine (depending on other rules regarding documentation).

What do you think?

romani: > while the latter does not. 

Second code is also not good for diff.
It should be:
```
enum E {
    ONE,
    TWO,
    THREE,
     ;
}
```

That is why I am looking at single message. As we validating single and simple pattern - comma is after value, and last symbol at the line.

Am I missing smth ?

kariem: > Second code is also not good for diff.

Yes, this is what I wrote.

The important feature is that you only change lines of code that are important for the change. If I add or remove one enum constant, only one line should be affected.

I understand the demand for a single message. I wanted to have two messages, one simple and one specific for the two cases. It's clear enough to write that there should be a trailing comma after every constant. If someone puts a semi-colon after the comma on the same line, then we should show the more specific message. If you think, only having one very specific message in there, please do not forget that specificity does not necessarily provide clarity. I always prefer the latter over the prior when we talk about error messages.

I also do not see a reason not to provide concise information for the user so that they know how to fix the violation. Is there a reason?

In addition, I do not think we should use the word "symbol" here. Especially as this code tries to enforce a level of consistency right above lexical correctness, we should not be to lax and use the correct terms. A comment is also a symbol, and both, traditional comments and end-of-line comments should not be considered a problem in the context of the check for trailing commas on enumerations. Hence, we should not use "last symbol" in the line, as there are actually two permitted symbols after the comma.

``` java
enum E {
    ONE,
    TWO,
    THREE, // this is ok
}
```

``` java
enum E {
    ONE,
    TWO, /* although confusing, this is also ok */
    THREE, 
}
```

``` java
enum E {
    ONE, /* although this does not make any sense, */ // this is also ok
    TWO, 
    THREE, 
}
```

@rnveach @romani, Please decide on either using one very specific message, or two concise and clear messages.

One message:

    Enum constant should be followed by only a trailing comma in the line.

We will still have to work on the wording, because, as you see in the examples above, this is not correct when we have comments on the same line after the enum constant (and trailing comma).

Two messages:

    Enum constant should be followed by a trailing comma in the line.
    Enum constant should be followed by a trailing comma in the line and no additional semicolon.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: Can we mark this as resolved?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
romani: > I also do not see a reason not to provide concise information for the user so that they know how to fix the violation. Is there a reason?

sorry that I did not shared a reason ... in sevntu project it is not a issue. But ones your Check start its way to main library .... each message need to be translated, each plugin need to probably register it and so on, so we try to avoid extra messages.

Looks like we need to vote. As there are three of us. 
@rnveach , please share your vote, and looks like your vote actually made a decision :).

> We will still have to work on the wording, because, as you see in the examples above, this is not correct when we have comments on the same line after the enum constant

I do not see big problem with this, almost all checks ignore comments completely.  I like your message.

kariem: Done: 6c53368c49d477916945c1e69ee3c6f50546f78e

kariem:
@romani @rnveach I have consolidated the messages in 6c53368c49d477916945c1e69ee3c6f50546f78e and then squashed everything again. See c48049f52a763e317277124f42761c73c5b6a3fa

Is there anything else I can do for this PR to be closed?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: Also, I have extracted the actual check to a separate method, as [previously suggested by @romani](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/709#discussion_r217871003).

eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.xml
```diff
@@ -556,5 +556,13 @@
       <message-key key="require.fail"/>
     </rule-metadata>
 
+    <rule-metadata name="%EnumTrailingCommaCheck.name" internal-name="EnumTrailingCommaCheck" parent="TreeWalker">
+      <alternative-name internal-name="com.github.sevntu.checkstyle.checks.coding.EnumTrailingCommaCheck"/>
+      <description>%EnumTrailingCommaCheck.desc</description>
+
+      <message-key key="enum.trailing.comma"/>
+      <message-key key="enum.trailing.comma.semi"/>
```
romani: please remove this line.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
```diff
@@ -0,0 +1,162 @@
...
+            final DetailAST lastComma = getLastComma(constant);
+
+            final DetailAST nextAst = lastComma.getNextSibling();
+            if (isIllegalTokenAfterComma(lastComma, nextAst)) {
+                log(nextAst, MSG_KEY);
+            }
+        }
+    }
+
+    private boolean isIllegalTokenAfterComma(DetailAST lastComma, DetailAST nextAst) {
```
romani: please explain why this method is not `isSmthAfterCommaOnSameLine` with no special processing of SEMI.
I think it should be checking for any sibling on same line.

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java
```diff
@@ -36,7 +36,10 @@
 
     /** List of checks to suppress if we dynamically add it to the configuration. */
     private static final List<String> ADD_CHECK_SUPPRESSIONS = Arrays
-            .asList("ReturnCountExtendedCheck");
+            .asList(
+                    "EnumTrailingCommaCheck",
```
romani: please comment above "// conflicting with other Checks"

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java
```diff
@@ -36,7 +36,10 @@
...
+                    "ReturnCountExtendedCheck"
```
romani: please add comment above "// initial ",
we will update it ones your PR is merged.
@rnveach , do you remember a reason ?

rnveach:
CI is failing.
https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/jobs/442062494#L559
> [ERROR] src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java:[134,5] (javadoc) JavadocMethod: Missing a Javadoc comment.
[ERROR] src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java:[134,5] (coding) ReturnCount: Return count is 2 (max allowed for non-void methods/lambdas is 1).

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java
rnveach: @romani What was the other check in conflict here? We should probably specifically list it.

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java
rnveach: @romani The reason was probably because it wasn't enabled in main repo as this is regression script.
See https://github.com/checkstyle/checkstyle/issues/3496 where it was enforced.
This line can probably be removed.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: According to the Java Language Specification on [Enum Constants](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.9.1) the only allowed symbols are

1. Other enum constants, separated by a comma
2. Enum body declarations, starting with a semicolon

If this method does not allow any sibling on the same line, this would also include traditional comments (starting with `TokenType.BLOCK_COMMENT_BEGIN`) and end-of-line comments (`TokenType.SINGLE_LINE_COMMENT`). Both, however, should be allowed. Am I misunderstanding something?

eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.xml
kariem: Done: 5ee1a62b17cf9336b1d2558c6007721077492dae

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java
kariem: Done: 038ff42971b6e8c9ebefb28e13c9a5680ef66091

Added the comment directly to the line of `EnumTrailingCommaCheck` and included the checks in conflict as already outlined in [this comment on checkstyle issue 6054](https://github.com/checkstyle/checkstyle/pull/6054#issuecomment-409405662).

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java
kariem: See [my comment on the above review note](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/709#discussion_r225831838): I have only added the comment to the line for `EnumTrailingCommaCheck`. I don't think touching the code regarding `ReturnCountExtendedCheck` is in the scope of this PR.

kariem:
> CI is failing. https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/jobs/442062494#L559

Resolved in 26d7bf9c4e9432d90f3e39cab0edb6192abae59f and bc2730a22bec2ae03924ee1e109abbcd1dffffff.

kariem:
Squashed all changes and pushed again: 4a30256551bb9cac189e3f6dd592e19143cb064e

Please let me know what you think needs to be done in addition to merge this pull request.

It would be very helpful, if you could write all up in one go. The back and forth is very time consuming for all parties involved.

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java
romani: ok, look ok now , I will update reason for second issue , ones we merge you.

romani:
> It would be very helpful, if you could write all up in one go.

Sorry, it is all my bad, I agree, but I do review of all PRs and requests and issues, it is hard fro me to spend whole my time only on one PR and make it merged with minimal amount of interactions.
In most of the time, I just do not see all problems till the very end. Sorry my bad, I hope to improve my skill to see all conner cases in one review.
I appreciate your persistence to make this PR be merged.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: @romani Just checked the implementation again. Good to know that we do not need to check for comments, so the explicit check on symbols seems to be redundant.

However, after looking at this again, I still think the current version is very clear on what is done: we have two problem cases:

1. The last comma is followed by an enum constant: this means that there is no more comma after that enum constant, because we have already checked for the last comma

2. The last comma is followed by a semicolon on the same line.

This is the implementation

``` java
private boolean isIllegalTokenAfterComma(DetailAST lastComma, DetailAST nextAst) {
    boolean illegalToken = false;

    final int nextAstType = nextAst.getType();
    if (nextAstType == TokenTypes.SEMI) {
        if (nextAst.getLineNo() == lastComma.getLineNo()) {
            // semi on the same line as last comma
            illegalToken = true;
        }
    }
    else if (nextAstType == TokenTypes.ENUM_CONSTANT_DEF) {
        // next enum constant on the same line after comma
        illegalToken = true;
    }
    return illegalToken;
}
```

I have now tried different options and could not find a clearer solution, so I do not really understand what you are asking for when you do not want the "extra validatoin" of tokens. We will always have to check for the semicolon or the right curly bracket to see whether we are at the end of the enum constant block.

Also, this method could be a lot less verbose so I have just updated it. See be8be03107a778e9d91c70341a29fc7c5c21a356

``` java
private boolean isIllegalTokenAfterComma(DetailAST lastComma, DetailAST nextAst) {
    final int nextAstType = nextAst.getType();

    // semi on the same line as last comma, or followed by enum constant
    return (nextAstType == TokenTypes.SEMI && nextAst.getLineNo() == lastComma.getLineNo())
            || nextAstType == TokenTypes.ENUM_CONSTANT_DEF;
}
```

Does that work for you?

kariem:
Squashed all changes and pushed again: 1b8739896bd25b20d496882f8bcedce7a20869bb

Please let me know what you think needs to be done in addition to merge this pull request.

sevntu-checks/src/test/java/com/github/sevntu/checkstyle/internal/CheckstyleRegressionTest.java
kariem: Can we mark this part of the conversation as resolved?

rnveach:
@kariem there is still a checkstyle violation, be sure to run `mvn verify test` before pushing to the server
> [INFO] There is 1 error reported by Checkstyle 8.12 with sevntu-checks.xml ruleset.
[ERROR] src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java:[147,17] (extension) LogicConditionNeedOptimization: Condition with || at line 147 position 16 need optimization. All method calls are advised to move to end of logic expression.

kariem:
> there is still a checkstyle violation, be sure to run `mvn verify test` before pushing to the server

@rnveach I did run this and did not get any errors. I re-ran it again after an `mvn clean` and still did not get any errors.

> [ERROR] src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java:[147,17] (extension) LogicConditionNeedOptimization: Condition with || at line 147 position 16 need optimization. All method calls are advised to move to end of logic expression.

Thank you for the information. This relates to this line:

``` java
// semi on the same line as last comma, or followed by enum constant
return (nextAstType == TokenTypes.SEMI && nextAst.getLineNo() == lastComma.getLineNo())
        || nextAstType == TokenTypes.ENUM_CONSTANT_DEF;
```

As I cannot reproduce this and do not want to push changes until it works, could you tell me, how I can correct this line? Should I blow it up again as we had this as indicated in https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/709#discussion_r226859172 ?

Thank you for your help.

rnveach:
Sorry, for sevntu to check itself the command is `mvn -e verify -Pno-validations,selftesting`.

The violation just wants you to swap the contents between the `||`. So instead of `a || b`, it wants `b || a`.

rnveach:
Sorry, for sevntu to check itself the command is `mvn -e verify -Pno-validations,selftesting`.

The violation just wants you to swap the contents between the `||`. So instead of `a || b`, it wants `b || a`.

rnveach:
Sorry, for sevntu to check itself the command is `mvn -e verify -Pno-validations,selftesting`.

The violation just wants you to swap the contents between the `||`. So instead of `a || b`, it wants `b || a`.

rnveach:
Sorry, for sevntu to check itself the command is `mvn -e verify -Pno-validations,selftesting`.

The violation just wants you to swap the contents between the `||`. So instead of `a || b`, it wants `b || a`.

kariem:
Thank you for the clarification! Just updated this part.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/InputEnumTrailingCommaCheck.java
```diff
@@ -0,0 +1,61 @@
...
+    enum E4 {
+        ONE,
+        TWO,
+        THREE,; // violation
+    }
+
+    enum E5 {
+        ONE,
+        TWO,
+        THREE,
```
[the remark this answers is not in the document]
kariem: @romani 

Added a separate section at the end for both, traditional and end-of-line comments.

``` java
// check ignores comments
enum E9_1 {
    ONE, // this is ok
    TWO,
    THREE,
}
enum E9_2 {
    ONE,
    TWO /* this is also ok */ ,
    THREE,
    ;
}
```

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/EnumTrailingCommaCheck.java
kariem: I have rewritten that whole block. It's now similar to the ArrayTrailingCommaCheck, with some additional comments:

`enumConstEnd` can either be SEMI or RCURLY.

``` java
// Only check, if block is multi-line and there are more than one enum constants
if (enumConstLeft.getLineNo() != enumConstEnd.getLineNo()
        && enumConstBlock.getChildCount(TokenTypes.ENUM_CONSTANT_DEF) > 1) {

    final DetailAST lastSibling = enumConstEnd.getPreviousSibling();
    if (lastSibling.getType() != TokenTypes.COMMA) {
        // constant definition does not end in a comma
        log(lastSibling, MSG_KEY);
    } else if (enumConstEnd.getType() == TokenTypes.SEMI
            && enumConstEnd.getLineNo() == lastSibling.getLineNo()) {
        // constant definition ends in a comma, but is followed by semi on same line
        log(enumConstEnd, MSG_KEY);
    }
}
```

With this, we can remove the other helper methods `getLastComma` and `isIllegalTokenAfterComma`.

Does that work for you?

kariem: @romani I haven't directly answered the question, but I think the current solution is better than I had previously suggested. Let me quickly answer the question

>  Fact that I do not understand is why we care about "last comma"?

We only care about the "last comma". This is the point of the check. Hence the name Enum**TrailingComma**Check. Intermediary commas are required as per language specification. Only the last, the trailing comma, is relevant for this check.

kariem:
@romani From my perspective this is done. From what I can see, all the side-conversations that have not been closed or resolved are either resolved or outdated and redundant.

If there is anything else for me to do here, please write one comment with everything that needs to be done. This way I can do everything you want and have this pull request closed. The current process is very annoying and time consuming.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/InputEnumTrailingCommaCheck.java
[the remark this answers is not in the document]
kariem: @romani I missed to address the first part of this request.

Changes:

- Separate section with your exampe, slightly adapted
    ``` java
    // all ok
   enum E_1 {
       ONE, TWO
        , THREE, FOUR,
    }
    ```
    The trailing comma after `FOUR` is important here
- Included another check on the missing enum, if followed by a semicolon on the next line
    ``` java
    enum E2_2 {
        ONE,
        TWO,
        THREE  // violation
        ;
    }
    ````

I have also added more information to the `// violation` comments to explain the respective reasons.

romani:
@kariem , sorry that it takes so long but idea of Check somewhere is broken that is wh awe sendup with long discussion.
Lets close this PR and create new PR to start discussion from scratch.


I still see inconsistency in behavior and what Check try to address.
at javadoc 
>  Checks if enum constant contains an optional trailing comma.
Behavior: 
```

   enum Type {
       ALPHA,
       BETA,
       GAMMA,   // changed due to the ','
       DELTA    // new line
  }

could be done as
   enum Type {
       ALPHA,
       BETA,
       GAMMA   // no changes
       , DELTA    // new line
  }


     enum E_1 {
         ONE, TWO
         , THREE, FOUR,
     }
```

We need to clearly define what we are validating.

Se you in new PR.

kariem:
@romani Please describe what the inconsistency is for you and I will try to address this. The idea of the check follows [ArrayTrailingComma](http://checkstyle.sourceforge.net/config_coding.html#ArrayTrailingComma). If you do not want what is done in ArrayTrailingComma, we have a different discussion.

I do not think a new PR helps here, as you might have noticed, this is PR number 4 and I have addressed *all* your requests in the previous PRs and in this PR.

See #722 for a follow-up.

romani:
Javadoc first sentence just tell about training comma, but in fact Check care only about comma after last element. This need to be fixed in javadoc if this is intentional.

This check does not demand each element of enum to be on  separate line. If that is intentional we just need to make it clear in javadoc, to make it clear that this Check address only particular case.

romani:
Sorry that I ma demanding, but I do this for a reason. There are hundreds of Checks and their goals a not very clear that result in many bugs reported on Checks and had to keep project under control.

Please see my comments in new PR.

PR #720, opened 2018-09-19
Handle lambdas in ForbidCertainMethodCheck

skitt:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/721

ForbidCertainMethodCheck allows the number of parameters to be used to
specify forbidden methods more precisely than using only their name;
however it counds those parameters by counting EXPR tokens. Lambdas
appear as LAMBDA tokens, not EXPR tokens, so they aren’t counted
towards the real argument count, resulting in false-positives or
false-negatives depending on the test being run.

This patch changes the argument count to count commas instead, all at
the same level just under the ELIST. This ensures that arguments are
counted regardless of their token type.

Signed-off-by: Stephen Kitt <skitt@redhat.com>

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ForbidCertainMethodCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ForbidCertainMethodCheck.java
@@ -161,7 +161,7 @@ public class ForbidCertainMethodCheck extends AbstractCheck {
                 else {
                     methodNameInCode = dot.getLastChild().getText();
                 }
-                final int numArgsInCode = dot.getNextSibling().getChildCount(TokenTypes.EXPR);
+                final int numArgsInCode = getMethodCallParameterCount(ast);
                 if (isForbiddenMethod(methodNameInCode, numArgsInCode)) {
                     log(ast, MSG_KEY, methodNameInCode, methodName,
                         numArgsInCode, argumentCount);
@@ -171,6 +171,24 @@ public class ForbidCertainMethodCheck extends AbstractCheck {
                 Utils.reportInvalidToken(ast.getType());
                 break;
         }
+    }
+
+    /**
+     * Count the parameters given to a method call.
+     * @param ast The method call AST.
+     * @return The number of parameters.
+     */
+    private static int getMethodCallParameterCount(DetailAST ast) {
+        int paramCount = 0;
+        final DetailAST expressionList = ast.getFirstChild().getNextSibling();
+        // This works by counting the number of commas separating the
+        // expressions passed to the method, if any
+        if (expressionList.getChildCount() > 0) {
+            // We have at least one parameter, so the total number of
+            // parameters is the number of commas plus one
+            paramCount = expressionList.getChildCount(TokenTypes.COMMA) + 1;
+        }
+        return paramCount;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


skitt:
I’ll update this to follow the expected commit message format.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ForbidCertainMethodCheck.java
```diff
@@ -173,6 +173,20 @@ public void visitToken(DetailAST ast) {
         }
     }
 
+    /**
+     * Count the parameters given to a method call.
+     * @param expressionList The expression list containing the parameters.
+     * @return The number of parameters.
+     */
+    private int getMethodCallParameterCount(DetailAST expressionList) {
```
romani: getMethodCallParameterCount vs expressionList ... mismatch

please make this method to be based on method AST, like `getMethodCallParameterCount(ast)` , and please make  method as static, one dat it will be moved to some Util class.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ForbidCertainMethodCheck.java
```diff
@@ -173,6 +173,20 @@ public void visitToken(DetailAST ast) {
...
+        int paramCount = 0;
+        if (expressionList.getChildCount() > 0) {
+            // At least one parameter, count commas
```
romani: this comment is misleading.
please make it as "amount of parameters is always +1 of amount of commas" or remove this comment.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/InputForbidCertainMethodCheck.java
```diff
@@ -4,6 +4,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 package com.github.sevntu.checkstyle.checks.coding;
 
+import java.util.function.Supplier;
```
romani: this line caused a big changes in UTs expected results, please change example to avoid extra import or remove empty line between imports to keep code on the same line further down.

skitt:
I probably won’t have time to do the regression testing report before next week.

romani:
Code changes looks good, but let's wait for regression testing, we might need to get a bit more input cases from it.

>I probably won’t have time to do the regression testing report before next week.

No problem. Do not hesitate to ask us questions.

skitt:
> I probably won’t have time to do the regression testing report before next week.

It took me a while longer... The reports are up:

* https://skitt.github.io/reports/master/RxJava/site/
* https://skitt.github.io/reports/master/Vavr/site/
* https://skitt.github.io/reports/master/guava/site/
* https://skitt.github.io/reports/master/infinispan/site/
* https://skitt.github.io/reports/master/jOOL/site/
* https://skitt.github.io/reports/master/protonpack/site/

This was run with the ForbidCertainMethod check enabled (but no arguments).

romani:
Ok, thanks a lot for reminding

rnveach:
Here are the regression reports:

http://rveach.no-ip.org/checkstyle/regression/reports/187/

Default module showed no changes.

-------

http://rveach.no-ip.org/checkstyle/regression/reports/188/

All methods (using `.*` for property `methodName`) shows lambdas as arguments to a method call being reported as `0` argument methods are changed to `1` argument methods.
This behavior looks correct as these methods never had 0 arguments.

Some even had more than 1 argument.
http://rveach.no-ip.org/checkstyle/regression/reports/188/checkstyle/index.html#A17
http://rveach.no-ip.org/checkstyle/regression/reports/188/checkstyle/xref/src/main/java/com/puppycrawl/tools/checkstyle/PackageObjectFactory.java.html#L297

-------

http://rveach.no-ip.org/checkstyle/regression/reports/189/

I then ran one more regression with all methods and where argument count was 1+ to see if I saw anything different then other reports.
It looks to be the same as the previous report, mostly all new violations instead of a change in violation.

romani:
Tht a lot for report, now I am more confident  in fix

Issue #733, opened 2019-03-31
update Travis build item for eclipse compiler 


romani:
Problem: https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/jobs/513645584#L2157
```
/home/travis/.m2/repository/R-4.7-201706120950/ecj-4.7.jar is not found, downloading ...
--2019-03-31 02:26:23--  http://ftp-stud.fht-esslingen.de/pub/Mirrors/eclipse/eclipse/downloads/drops4/R-4.7-201706120950/ecj-4.7.jar
Resolving ftp-stud.fht-esslingen.de (ftp-stud.fht-esslingen.de)... 129.143.116.10, 2001:7c0:700::10
Connecting to ftp-stud.fht-esslingen.de (ftp-stud.fht-esslingen.de)|129.143.116.10|:80... connected.
HTTP request sent, awaiting response... 404 Not Found
2019-03-31 02:26:24 ERROR 404: Not Found.
[ERROR] Command execution failed.
org.apache.commons.exec.ExecuteException: Process exited with an error: 8 (Exit value: 8)
    at org.apache.commons.exec.DefaultExecutor.executeInternal (DefaultExecutor.java:404)
    at org.apache.commons.exec.DefaultExecutor.execute (DefaultExecutor.java:166)
```

solution is probably to reuse main project approach - https://github.com/checkstyle/checkstyle/blob/master/.ci/eclipse-compiler-javac.sh

rnveach:
Fix was merged

PR #735, opened 2019-04-07
Issue #733: re-enable and fix eclipse-compiler

rnveach:
Issue #733


rnveach:
travis item is passing and i don't see any recognizable differences from main repo's travis output

Issue #743, opened 2019-05-07
Fix errors when compiling with checkstyle 8.21


rnveach:
Identified at https://github.com/checkstyle/checkstyle/pull/6629#issuecomment-490267933

Main:
````
[ERROR] /sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/design/ForbidWildcardAsReturnTypeCheck.java:[308,60] incompatible types: com.puppycrawl.tools.checkstyle.api.DetailAST cannot be converted to antlr.collections.AST
[ERROR] /sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/DiamondOperatorForVariableDefinitionCheck.java:[86,40] cannot find symbol
  symbol:   method equalsTree(com.puppycrawl.tools.checkstyle.api.DetailAST)
  location: variable varDefArguments of type com.puppycrawl.tools.checkstyle.api.DetailAST
[ERROR] /sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ForbidCCommentsInMethodsCheck.java:[45,47] com.puppycrawl.tools.checkstyle.api.DetailAST is abstract; cannot be instantiated
````
First one is because it is using ANTLR class instead of DetailAST.
Second uses an ANTLR specific method, `equalsTree`.
Third is because it tries to instantiate DetailAST but it can't anymore since it is an interface.

rnveach:
Test code can't be fixed until we upgrade to new version.

rnveach:
tests moved to https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/745
All fixes done for this issue.

PR #744, opened 2019-05-07
Issue #743: fix compiling issues with checkstyle 8.21

rnveach:
Issue #743

Regression to come.


rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/248/

No differences.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/DiamondOperatorForVariableDefinitionCheck.java
```diff
@@ -113,4 +113,46 @@ private static DetailAST getFirstTypeArgumentsToken(DetailAST rootToken) {
...
+        boolean result;
+
+        if (isAstEqual(left, right)) {
+            result = true;
+
+            DetailAST leftChild = left.getFirstChild();
+            DetailAST rightChild = right.getFirstChild();
+
+            while (leftChild != rightChild) {
+                if (!isTreeEqual(leftChild, rightChild)) {
```
rnveach: There is no protection for one being `null` and the other not being `null`. Regression couldn't find any cases.
We check `type` when scrolling through the children, so for one to be `null` we would need a token that has an optional child. I couldn't come up with any cases.

Issue #751, opened 2019-06-19
Jsr305AnnotationsCheck is crashing on checkstyle sources


romani:
details at https://travis-ci.org/checkstyle/checkstyle/jobs/547533522#L415
```
Caused by: java.lang.NullPointerException
    at com.github.sevntu.checkstyle.checks.coding.Jsr305AnnotationsCheck$AbstractJsr305Handler.isPrimitiveType (Jsr305AnnotationsCheck.java:939)
    at com.github.sevntu.checkstyle.checks.coding.Jsr305AnnotationsCheck$ParameterJsr305Handler.runHandler (Jsr305AnnotationsCheck.java:603)
    at com.github.sevntu.checkstyle.checks.coding.Jsr305AnnotationsCheck$AbstractJsr305Handler.handle (Jsr305AnnotationsCheck.java:807)
    at com.github.sevntu.checkstyle.checks.coding.Jsr305AnnotationsCheck.visitToken (Jsr305AnnotationsCheck.java:371)
```

diff: https://github.com/checkstyle/checkstyle/pull/6836/files

command in checkstyle repo (with change): `./.ci/travis/travis.sh checkstyle-and-sevntu`

ATTENTION: to unblock checkstyle master build I reverted upgrade to sevntu 1.34.0 .
I will not do any public notifications on new senvtu release till issue is resolved. 

It is crashing not only on checkstyle sources: sevntu CI master build is red now - https://app.wercker.com/checkstyle/sevntu.checkstyle/runs/build/5d09aa43105780001c558e65?step=5d09aa88322dac0009d85364
null pointer at "Apache Struts" sources.

romani:
@mbert , please address ASAP.

romani:
wercker was green on PR, as it takes master contribution repo, where new Check did not, as PR is hanging, and it is failing right after merge .... ok.
So new Checks are skipped from such regression testing ....

romani:
https://travis-ci.org/checkstyle/checkstyle/jobs/547533522#L462
`Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing /home/travis/build/checkstyle/checkstyle/src/main/java/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.java`

This file passed during regresson testing:
https://mbert.github.io/sevntu.checkstyle/reports/diff/checkstyle/index.html#A3008
[...]

PR #752, opened 2019-06-19
Issue #751: Jsr305AnnotationsCheck is crashing on checkstyle sources

mbert:
This fixes the issue from #751.

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/Jsr305AnnotationsCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/Jsr305AnnotationsCheck.java
@@ -532,11 +532,78 @@ public class Jsr305AnnotationsCheck extends AbstractCheck {
     private static void addNextNullnessAnnotation(final Set<NullnessAnnotation> result,
             DetailAST ast) {
         final DetailAST identifier = ast.findFirstToken(TokenTypes.IDENT);
-        final String annotationName = identifier.getText();
-        final NullnessAnnotation annotation = STRING2ANNOTATION.get(annotationName);
-        if (annotation != null) {
-            result.add(annotation);
+        if (identifier != null) {
+            final String annotationName = identifier.getText();
+            final NullnessAnnotation annotation = STRING2ANNOTATION.get(annotationName);
+            if (annotation != null) {
+                result.add(annotation);
+            }
         }
+    }
+
+    /**
+     * Is the current symbol a primitive type.
+     * @param ast
+     *        the ast
+     * @return true if yes
+     */
+    private static boolean isPrimitiveType(DetailAST ast) {
+        boolean result = false;
+        final DetailAST parameterType = ast.findFirstToken(TokenTypes.TYPE);
+        final DetailAST identToken = parameterType.getFirstChild();
+
+        if (identToken != null) {
+            switch (identToken.getType()) {
+                case TokenTypes.LITERAL_BOOLEAN:
+                case TokenTypes.LITERAL_INT:
+                case TokenTypes.LITERAL_LONG:
+                case TokenTypes.LITERAL_SHORT:
+                case TokenTypes.LITERAL_BYTE:
+                case TokenTypes.LITERAL_CHAR:
+                case TokenTypes.LITERAL_VOID:
+                case TokenTypes.LITERAL_DOUBLE:
+                case TokenTypes.LITERAL_FLOAT:
+                    result = !isArrayOrElipsis(parameterType);
+                    break;
+                default:
+            }
+        }
+        return result;
+    }
+
+    /**
+     * Checks whether token is array or elipsis.
+     *
+     * @param identToken
+     *        the token
+     * @return true if yes
+     */
+    private static boolean isArrayOrElipsis(final DetailAST identToken) {
+        final DetailAST next = identToken.getNextSibling();
+        final boolean result;
+        switch (next.getType()) {
+            case TokenTypes.ARRAY_DECLARATOR:
+            case TokenTypes.ELLIPSIS:
+                result = true;
+                break;
+            default:
+                result = false;
+        }
+        return result;
+    }
+
+    /**
+     * Is the current symbol of void type.
+     * @param ast
+     *        the ast
+     * @return true if yes
+     */
+    protected static boolean isVoid(DetailAST ast) {
+        final DetailAST parameterType = ast.findFirstToken(TokenTypes.TYPE);
+        boolean result = false;
+        final DetailAST identToken = parameterType.getFirstChild();
+        result = identToken.getType() == TokenTypes.LITERAL_VOID;
+        return result;
     }
 
     /**
@@ -600,7 +667,8 @@ public class Jsr305AnnotationsCheck extends AbstractCheck {
             checkContainsAll(MSG_PARAM_NONNULL_AND_NULLABLE,
                     NullnessAnnotation.NONNULL, NullnessAnnotation.NULLABLE);
 
-            if (isPrimitiveType()) {
+            final DetailAST ast = getAst();
+            if (isPrimitiveType(ast)) {
                 checkContainsAny(MSG_PRIMITIVES_WITH_NULLNESS_ANNOTATION,
                         NullnessAnnotation.CHECK_FOR_NULL, NullnessAnnotation.NONNULL,
                         NullnessAnnotation.NULLABLE);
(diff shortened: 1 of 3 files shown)


mbert:
@rnveach @romani 
This fixes the immediate two issues we know of now:
1. NPE in `Jsr305AnnotationsCheck.java:939` when processing `SuppressWithPlainTextCommentFilter.java` from the checkstyle sources
2. NPE in `Jsr305AnnotationsCheck.java:535` when processing an unknown source file from the apache struts repo

For (1) creating a test case was easy. For (2), since I don't know which source is affected (do you know? did I miss something?) I have created a reflection-based unit test.

Why did this happen? As already explained in #752, the old code used guards whenever dealing with the result of one of the "find" and "get" methods from `DetailAST`. However in some cases it looked like those results would never actually become null, so that those guards would never fire in the unit tests. Subsequently, as discussed in the original PR, I removed some of those guards. 

I think that this approach is too dangerous, and the lesser evil would be to restore those guards and cover them in a reflection-based test. For cases like this IMO it's better to have one guard too many than any missing. 

Do you agree to my proposal? If yes, I'd proceed as proposed and finish this PR by the end of today.

mbert:
Replying to myself here. I have identified a smaller number of code locations I feel concerned about, re-added nullness guards and implemented one reflection-based unit test that covers them. 

To me this looks like a feasible solution to the problem. It's good we stumbled across this problem so quickly. With my proposed solution I feel much easier than before. Would this be acceptable for you?

romani:
Thanks a lot for fix, Please remove extra null checks that covered only by reflection tests. It is very error prone approach.

Please run regression testing on projects of all your accessible sources, no need to share report, so include even close sources. Such approach usually find almost all cases, very rare cases will be reported by users.

romani:
>(2), since I don't know which source is affected (do you know? did I miss something?)

Path file is in wercker log, here a copy:
`Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Exception was thrown while processing /pipeline/source/sevntu-checks/.ci-temp/contribution/checkstyle-tester/src/main/java/apache-struts/plugins/bean-validation/src/test/java/org/apache/struts/beanvalidation/actions/FieldMatchAction.java
    at com.puppycrawl.tools.checkstyle.Checker.processFiles (Checker.java:304)`

mbert:
> > (2), since I don't know which source is affected (do you know? did I miss something?)
> 
> Path file is in wercker log, here a copy:
Thank you, I'll pursue this, add the code that triggered the problem to the test data set and re-upload.

romani:
Correction: for regression testing please launch `launch.sh` or `launch.groovy` from https://github.com/checkstyle/contribution/tree/master/checkstyle-tester with your Check in config, and put severity as ignore on it. In such case it will report only exceptions, so you can run on massive amount of projects, and report will be small, ideally no errors (exceptions) in report.

mbert:
> Correction: for regression testing please launch `launch.sh` or `launch.groovy` from https://github.com/checkstyle/contribution/tree/master/checkstyle-tester with your Check in config, and put severity as ignore on it. In such case it will report only exceptions, so you can run on massive amount of projects, and report will be small, ideally no errors (exceptions) in report.

I have now finished the fix. No more NPEs were found using checkstyle tester. To make sure I ran the tester on the old code and was able to reproduce the problem that was there before. However I found that with some projects in `projects-to-test-on.properties` (I uncommented pretty much all of them) the build bails out with compilation errors or in some other cases with exceptions not related to the new check. The vast majority of projects from that file was fine. 

I'll check for comments / mails when home.

rnveach:
Sevntu CI also runs regression of newly introduced checks. It should be looked into why it didn't fail as it automatically runs checks on checkstyle source.

romani:
I explained reason of missing exception at https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/753#discussion_r295648282.
Right now Check is not new, so CI found regression quickly, as it is green now, it means problem is resolved.

romani:
@rnveach , please review PR.

romani:
tested on my local, no exception found.

@mbert , thanks a lot.
@rnveach , please finalize review.

rnveach:
@mbert I cannot reproduce exceptions using our regression tools on old code before this fix.

http://rveach.no-ip.org/checkstyle/regression/reports/252/
b56b4fa21e69b31cac6474ac844c18a58ee408ae 
This was the config you specified to use from contribution with the default values.
````
        <module name="com.github.sevntu.checkstyle.checks.coding.Jsr305AnnotationsCheck">
          <property name="severity" value="ignore"/>
          <property name="id" value="configured"/>
          <property name="packages" value="com,org,net"/>
          <property name="allowOverridingReturnValue" value="false"/>
          <property name="allowOverridingParameter" value="false"/>
        </module>
````


Am I doing something wrong?

**Comment by mbert** [comment]
> @mbert I cannot reproduce exceptions using our regression tools on old code before this fix.
> 
> http://rveach.no-ip.org/checkstyle/regression/reports/252/
> commit: [b56b4fa](https://github.com/sevntu-checkstyle/sevntu.checkstyle/commit/b56b4fa21e69b31cac6474ac844c18a58ee408ae)
> This was the config you specified to use from contribution with the default values.
> 
> ```
>         <module name="com.github.sevntu.checkstyle.checks.coding.Jsr305AnnotationsCheck">
>           <property name="severity" value="ignore"/>
>           <property name="id" value="configured"/>
>           <property name="packages" value="com,org,net"/>
>           <property name="allowOverridingReturnValue" value="false"/>
>           <property name="allowOverridingParameter" value="false"/>
>         </module>
> ```
> 
> Am I doing something wrong?

I used the following configuration (my `packages` did not contain `net`, but that was enough to reproduce the error in apache-struts):

```
        <module name="com.github.sevntu.checkstyle.checks.coding.Jsr305AnnotationsCheck">
          <property name="severity" value="ignore"/>
          <property name="packages" value="com,org"/>
          <property name="allowOverridingReturnValue" value="false"/>
          <property name="allowOverridingParameter" value="false"/>
        </module>
```

I then edited the file `projects-to-test-on.properties`, so that it would actually check apache-struts, i.e. uncomment line 35 (for my regression test after that I uncommented almost all lines). 

I ran `mvn -Pno-validations install` from my `sevntu-checks` directory, edited the `pom.xml` file in my `checkstyle-tester` directory setting `sevntu-checkstyle.version` to the version in `sevntu-checks` and ran the following command:

```
groovy diff.groovy --localGitRepo /path/to/my/checkstyle --patchBranch master --patchConfig checks-sevntu-error.xml --listOfProjects projects-to-test-on.properties --mode single
```

I ran straight into the NPE from the travis build in issue #751.

**Comment by rnveach** [comment]
I got it working, the issue was not enabling executeIgnoredModules .

File: sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/Jsr305AnnotationsCheck.java [hunk 1]
```diff
@@ -532,11 +532,78 @@ private AbstractJsr305Handler handleDefinition(final DetailAST ast) {
...
+        }
+    }
+
+    /**
+     * Is the current symbol a primitive type.
+     * @param ast
+     *        the ast
+     * @return true if yes
+     */
+    private static boolean isPrimitiveType(DetailAST ast) {
```
rnveach: Why did these methods move from `AbstractJsr305Handler`? I don't see them needing anything from the class they were moved to.
Why not just move these methods over the locations they were originally?

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/InputJsr305AnnotationsCheckWithNestedAnnotation.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
rnveach: I don't think we need this license in the input file.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/Jsr305AnnotationsCheck.java
mbert: I did that when implementing the (subsequently removed) reflection-based unit test and left them there afterwards because I usually like static stuff better than non-static stuff. I'd like to leave it as it is now, but if you request it I can move it back.

sevntu-checks/src/test/resources/com/github/sevntu/checkstyle/checks/coding/InputJsr305AnnotationsCheckWithNestedAnnotation.java
mbert: Ah, copy-and-paste devil. Done.

romani:
>the issue was not enabling executeIgnoredModules .

Where do you set it ?

romani:
@rnveach , it would be good to finish review sooner .

rnveach:
> > the issue was not enabling executeIgnoredModules .

> Where do you set it ?

https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/launch_diff.sh#L212-L213

PR #753, opened 2019-06-19
config: fix sevntu regression

rnveach:
Looking for a failure.

### DIFF
--- a/.ci/travis.sh
+++ b/.ci/travis.sh
@@ -75,14 +75,17 @@ all-sevntu-checks-contribution)
 
 checkstyle-regression)
   git clone https://github.com/checkstyle/checkstyle
+  # update checkstyle_sevntu_checks.xml file in checkstyle for new modules
   cd sevntu-checks
   mvn -e install -Pno-validations
   mvn -e test -Dtest=CheckstyleRegressionTest#setupFiles -Dregression-path=../
   cd ../
+  # execute checkstyle validation on updated config file
   cd checkstyle
   mvn -e clean verify -e -DskipTests -DskipITs -Dpmd.skip=true \
       -Dfindbugs.skip=true -Dcobertura.skip=true \
-      -Dmaven.sevntu-checkstyle-check.checkstyle.version=8.12
+      -Dmaven.sevntu-checkstyle-check.checkstyle.version=8.18 \
+      -Dmaven.sevntu.checkstyle.plugin.version=1.34.0
   ;;
 
 eclipse-analysis)


rnveach:
Now that I think of this more, regression won't fail since it uses default configuration and this check produces no violations by default, it does nothing basically.

.ci/travis.sh
```diff
@@ -82,7 +82,8 @@ checkstyle-regression)
   cd checkstyle
   mvn -e clean verify -e -DskipTests -DskipITs -Dpmd.skip=true \
       -Dfindbugs.skip=true -Dcobertura.skip=true \
-      -Dmaven.sevntu-checkstyle-check.checkstyle.version=8.12
+      -Dmaven.sevntu-checkstyle-check.checkstyle.version=8.18 \
+      -Dmaven.sevntu.checkstyle.plugin.version=1.34.0
```
rnveach: I believe these values are correct. We should change their values to not be hard coded to prevent any issues.

romani: > maven.sevntu-checkstyle-check.checkstyle.version

it is the same value as in checkstyle repo - https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L206
as we clone latest, no need to override.

> maven.sevntu.checkstyle.plugin.version

no need it to be overridden to released version, it should be the same version as we have in checkstyle and version that we have in sevntu repo (as we substitute it by local build) https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/pom.xml#L14
I bumped it at https://github.com/checkstyle/checkstyle/pull/6838 , it should pass, as check for new senvtu Checks do not consider sevntu version, it just get list of Checks from html in web(latest leased state).

=====

Reason of problem is that Check config https://github.com/checkstyle/contribution/blob/master/checkstyle-tester/checks-sevntu-error.xml#L211 (that is ok by target) was not merged, so it was not present in master of contribution, wercker do use only master state of configs.
New modules are out of scope for regression testing, it was always fine as we require it during acceptance.
Problem was caused by fact that we did not ask regression report right before merge, and latest report was not done on our regular projects, but was done on projects where target annotations are in use. So it give us some sort of illusion that general project are passed and fixing only doc and minors.

romani: lets just remove this override, I am not sure what is a reason of it.

rnveach: > lets just remove this override, I am not sure what is a reason of it.
> New modules are out of scope for regression testing

This is dynamic regression testing. It will automatically take checkstyle's sevntu config and add any missing modules not listed. This regression testing didn't pick up anything because:
1) it was using the wrong sevntu version
2) The new module did no violation by default. It has to be configured to create violations and this dynamic regression only adds the default module.

> it is the same value as in checkstyle repo 
> should be the same version as we have in checkstyle and version that we have in sevntu repo 

Because there is no guarantee that can be the case, we should keep these values here and not hard code them but specifically use the values from our own sevntu POMs. If the 2 ever get out of sync again by mistake, regression will do nothing.

romani:
code was extended with comments to make it more clear of how regression happening

romani:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/wiki/How-to-release-sevntu.checkstyle was updated to mention that such version should be updated during release process.

.ci/travis.sh
romani: thanks a lot for explanation, I updated release process and added comments to code, as I am not sure I will remember this in a year.

PR #755, opened 2019-06-22
config: remove hardcoded values for checkstyle regression

rnveach:
Fix regression to remove hardcoded values and rely on values defined in our POM. If these values were ever out of synch, checkstyle regression would not use PR version of sevntu and fall back on released version.

This is why we had the failure at https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/jobs/549161973 after the PR was merged. 1.34.1 was released and hardcoded values were not updated.

### DIFF
--- a/.ci/travis.sh
+++ b/.ci/travis.sh
@@ -77,6 +77,13 @@ checkstyle-regression)
   git clone https://github.com/checkstyle/checkstyle
   # update checkstyle_sevntu_checks.xml file in checkstyle for new modules
   cd sevntu-checks
+  SEVNTU_VERSION=$(mvn -e -q -Dexec.executable='echo' -Dexec.args='${project.version}' \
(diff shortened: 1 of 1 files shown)


rnveach:
https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/jobs/549167164#L280
> sevntu version:1.34.1
eclipse-cs version:8.18

Looks like values are being retrieved correctly.

https://travis-ci.org/sevntu-checkstyle/sevntu.checkstyle/jobs/549167164#L597
> Downloading from nexus-codehaus-snapshot: https://oss.sonatype.org/content/repositories/codehaus-snapshots/com/github/sevntu-checkstyle/sevntu-checkstyle-maven-plugin/1.34.1/sevntu-checkstyle-maven-plugin-1.34.1.pom

We probably should eventually add `sevntu-checkstyle-maven-plugin` install to the process, but for now it isn't hurting anything using released version. We have never truly updated the plugin anyways.

Issue #758, opened 2019-07-03
sevntu-checks should not bundle checkstyle library


rnveach:
Discussed at https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/746#issuecomment-507911189 ,

We saw that bundling checkstyle into sevntu-sonar plugin caused issues as sevntu-sonar has to be installed side by side with checkstyle-sonar. The 2 sonar plugins could have different versions of checkstyle and this caused conflicts.

But why does sevntu-checks bundle checkstyle? It seems odd that sevntu-sonar plugin should not bundle it but normal sevntu-checks should.
In our own checkstyle documentation, we mention that 3rd party checks should be appended with main checkstyle jar, so I wouldn't expect others to also embed the checkstyle JAR.
See https://checkstyle.sourceforge.io/cmdline.html#Usage_by_Classpath_update

>    Run checkstyle with custom module(s) (Root module, Checks, etc) in configuration file:
>    java -classpath MyCustom.jar;checkstyle-8.22-all.jar

We should look into if we can remove checkstyle from sevntu-checks, what this means, and if there are any side effects.

romani:
@rdiachenko , please help to fix this

rdiachenko:
Hmm.. that was easier than I thought.. Looks like sevntu-checks doesn't bundle checkstyle library.
```
~/work/workspace/sevntu.checkstyle/sevntu-checks [master|✔] $ ls target/
apidocs                        generated-sources       sevntu-checks-1.34.1.jar
checkstyle-cachefile           generated-test-sources  sevntu-checks-1.34.1-javadoc.jar
checkstyle-checker.properties  javadoc-bundle-options  sevntu-checks-1.34.1-sources.jar
checkstyle-checker.xml         maven-archiver          site
checkstyle-result.xml          maven-status            surefire-reports
classes                        pmd-main.xml            test-classes
cobertura                      pmd-test.xml
generated-classes              pmd.xml
```
```
✔ 22:01 ~/work/workspace/sevntu.checkstyle/sevntu-checks [master|✔] $ zipinfo target/sevntu-checks-1.34.1.jar | grep 'lib/.*jar'
✘-1 22:02 ~/work/workspace/sevntu.checkstyle/sevntu-checks [master|✔] $ zipinfo target/sevntu-checks-1.34.1.jar
Archive:  target/sevntu-checks-1.34.1.jar
Zip file size: 195748 bytes, number of entries: 104
[...]

PR #765, opened 2019-07-08
Issue #758: make checkstyle dependency as provided in sevntu-checks

rdiachenko:
Issue #758.

Checkstyle dependency was marked as provided in sevntu-checks module. Redundant exclusion was removed from sevntu-sonar plugin.

See comments about checking other extensions: [sevntu-checkstyle-idea-extension](https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/758#issuecomment-509392019), [sevntu-checkstyle-maven-plugin](https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/758#issuecomment-509404538), [eclipsecs-sevntu-plugin](https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/758#issuecomment-509401382).

### DIFF
--- a/sevntu-checks/pom.xml
+++ b/sevntu-checks/pom.xml
@@ -37,6 +37,7 @@
       <groupId>com.puppycrawl.tools</groupId>
       <artifactId>checkstyle</artifactId>
       <version>${checkstyle.eclipse-cs.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>com.puppycrawl.tools</groupId>
--- a/sevntu-checkstyle-sonar-plugin/pom.xml
+++ b/sevntu-checkstyle-sonar-plugin/pom.xml
@@ -41,12 +41,6 @@
       <groupId>com.github.sevntu-checkstyle</groupId>
       <artifactId>sevntu-checks</artifactId>
       <version>${project.version}</version>
-      <exclusions>
-        <exclusion>
-          <groupId>com.puppycrawl.tools</groupId>
-          <artifactId>checkstyle</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
   </dependencies>
 


sevntu-checks/pom.xml
```diff
@@ -37,6 +37,7 @@
       <groupId>com.puppycrawl.tools</groupId>
       <artifactId>checkstyle</artifactId>
       <version>${checkstyle.eclipse-cs.version}</version>
+      <scope>provided</scope>
```
rnveach: What does this mean to maven-checkstyle-plugin that included sevntu and never overrode the plugin's version of checkstyle?
This means that it will default to `maven-checkstyle-plugin`'s default checkstyle version where as before sevntu overrode the checkstyle version itself. Right?

> eclipsecs-sevntu-plugin doesn't include checkstyle jar
> sevntu-checkstyle-maven-plugin doesn't include checkstyle jar
> sevntu-checkstyle-idea-extension doesn't include checkstyle jar

If these don't include it, where does it get it's version of checkstyle from? When I looked at the POM's in eclipse, it listed checkstyle as a dependency from sevntu, but I don't see it in the created jars.
Is it like sonar and expects main checkstyle plugin to also be installed already?

> sevntu-checkstyle-maven-plugin

@romani How come we never used this plugin in main repo?

romani: https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L554

rdiachenko: @rnveach ,
> What does this mean to maven-checkstyle-plugin that included sevntu and never overrode the plugin's version of checkstyle?

It means that default checkstyle version will be used from `maven-checkstyle-plugin`.

> This means that it will default to maven-checkstyle-plugin's default checkstyle version where as before sevntu overrode the checkstyle version itself. Right?

Correct. When checkstyle dependency was of `compile` scope within sevntu-checks, that checkstyle was used inside `sevntu-checkstyle-maven-plugin` (because of the maven dependency mediation). With checkstyle dependency of `provided` scope, checkstyle from `maven-checkstyle-plugin` will be used instead.

> If these don't include it, where does it get it's version of checkstyle from?

For `sevntu-checkstyle-maven-plugin` see the answer above. For `sevntu-checkstyle-idea-extension` and `eclipsecs-sevntu-plugin` checkstyle was taken from the main plugins.

> Is it like sonar and expects main checkstyle plugin to also be installed already?

Yes, all extensions can not be used without their corresponding main plugins.

rnveach: Thanks. I am not familar with how they worked so thats why I was asking.
I was also looking for how the plugins would be affected by this removal that may change functionality.

So `maven-checkstyle-plugin` will be affected if they don't override checkstyle version.
Our plugins will remain the same because they can't work without checkstyle version.

rnveach:
@romani Since you gave approval before response, please see summary at https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/765#discussion_r302007777 of the effects of this change if merged.
If you are ok, feel free to merge.

romani:
> So maven-checkstyle-plugin will be affected if they don't override checkstyle version.

@rdiachenko , please update issue description and put MIGRATION NOTE to users be aware what to do after they use new version of sevntu.
Please let me know if you think it is good time to make release of sevntu.

rdiachenko:
@romani 
> @rdiachenko , please update issue description and put MIGRATION NOTE to users be aware what to do after they use new version of sevntu.

The last comment was updated: https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/758#issuecomment-509406574

> Please let me know if you think it is good time to make release of sevntu.

Yes, let's do it

PR #770, opened 2019-07-27
minor: rename misleading property

pzygielo:
As discussed below, I believe this property is used only to control version of CS itself.

### DIFF
--- a/.ci/travis.sh
+++ b/.ci/travis.sh
@@ -72,7 +72,7 @@ checkstyle-regression)
                    --non-recursive org.codehaus.mojo:exec-maven-plugin:1.3.1:exec)
   echo sevntu version:$SEVNTU_VERSION
   ECLIPSE_CS_VERSION=$(mvn -e -q -Dexec.executable='echo' \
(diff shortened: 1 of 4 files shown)


pzygielo:
> @pzygielo Github is marking this as a draft.

I did it.

> I also don't feel `newprop` is a better name for this option. 
It's not, hence draft.

> `eclipse-cs` clearly tells us it is the version number for `eclipse-cs` which we rely on.

It's not.

pzygielo:
I might not see it, but I think it's only here https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/9765a3bf0f02272f921aec264667328f2ca60339/.ci/travis.sh#L17 that eclipse-cs is used, and the version is hardcoded anyway.

In other cases it's simply CS version. At least as I see it, which might be wrong of course.

pzygielo:
Closing due to [comment](https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/771#discussion_r307965575)
> New property name is very misleading

Got it.

Issue #705, opened 2018-07-01
update release proccess to skip version bumps at eclipsecs-sevntu-plugin


romani:
eclipsecs-sevntu-plugin and  eclipsecs-sevntu-plugin-feature are depend on eclipse-cs version and checkstyle version in eclipse-cs.

if checkstyle project introduce compile breaking changes in his internals, sevntu-checks need to be recompiled. It will work for all all except for eclipsecs-sevntu-plugin as it need for eclipse-cs to be upgraded first to new checkstyle version.
Eclipse-cs is not very quick project in updates, so it block all other sevntu extension to stay on old version (sonar, maven,...). No updates for sonar and maven cause more problems for users as they can not use latest sonar plugin with latest sevntu any more, so users must stay on old versions of checkstyle in Sonar and maven if they use sevntu.

Even more it create long delay in activation of disablements like https://github.com/checkstyle/checkstyle/commit/d8b8832caecc2f30e02d4a4014e9dce1ea9cfdba in main repo, so make unwanted problems to plugins be not noticed ([example](https://github.com/checkstyle/checkstyle/issues/5985)). 

We need to have ability update checkstyle version in all except for eclipsecs-sevntu-plugin, eclipsecs-sevntu-plugin should continue to use old version of sevnu-checks and be consistent with eclispe-cs project.

romani:
it goes it worse:
`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (sevntu-checkstyle-check) on project checkstyle-sonar-plugin: Failed during checkstyle configuration: cannot initialize module TreeWalker - Cannot set property 'blockTypes' to 'LITERAL_IF,LITERAL_ELSE,LITERAL_WHILE,LITERAL_DO,                              LITERAL_FOR,LITERAL_SWITCH,LITERAL_TRY,LITERAL_CATCH' in module ChildBlockLength: InvocationTargetException: com/puppycrawl/tools/checkstyle/utils/TokenUtils: com.puppycrawl.tools.checkstyle.utils.TokenUtils -> [Help 1]`

now I have problem to build checkstyle-sonar-plugin.

rnveach:
> All we need is to update

It is not that easy. Yes, this is one of the things we must do in this issue, but this will affect the release.sh and the release notes.
[...]

PR #771, opened 2019-07-27
Upgrade CS

pzygielo:
Mandatory to understand and do:
0) Issue you are trying to fix/resolve has to have "approved" label.
1) Put in description of Pull Request reference to issue if it exists. Example: "Issue: #XXXXXX"
2) Commit message should adhere to the following rules:
   a) Must match one of the following patterns:\n"
      ^Issue #\\d+: .*$
      ^Pull #\\d+: .*$
      ^(minor|config|infra|doc|spelling): .*$
   b) It contains only one line of text
   c) Must not end with a period, space, or tab

To avoid multiple iterations of fixes and CIs failures, please read http://checkstyle.sourceforge.net/contributing.html

ATTENTION: We are not merging Pull Requests that not passing our CIs, but we help to resolve issues.

Thanks for reading, remove whole this message and type what you need.

### DIFF
--- a/sevntu-checks/pom.xml
+++ b/sevntu-checks/pom.xml
@@ -16,8 +16,8 @@
   <properties>
     <project.build.sourceEncoding>iso-8859-1</project.build.sourceEncoding>
     <!-- It is compile dependency to checkstyle, this version has to be the same as eclipse-cs depends on -->
-    <checkstyle.eclipse-cs.version>8.18</checkstyle.eclipse-cs.version>
-    <checkstyle.configLocation>https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-${checkstyle.eclipse-cs.version}/config/checkstyle_checks.xml</checkstyle.configLocation>
+    <checkstyle.version>8.18</checkstyle.version>
+    <checkstyle.configLocation>https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-${checkstyle.version}/config/checkstyle_checks.xml</checkstyle.configLocation>
     <checkstyle.plugin.version>3.0.0</checkstyle.plugin.version>
     <sevntu.maven.plugin>RELEASE</sevntu.maven.plugin>
     <maven.pmd.plugin.version>3.11.0</maven.pmd.plugin.version>
@@ -36,7 +36,7 @@
     <dependency>
       <groupId>com.puppycrawl.tools</groupId>
       <artifactId>checkstyle</artifactId>
-      <version>${checkstyle.eclipse-cs.version}</version>
+      <version>${checkstyle.version}</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
@@ -44,7 +44,7 @@
       <artifactId>checkstyle</artifactId>
       <type>test-jar</type>
       <scope>test</scope>
-      <version>${checkstyle.eclipse-cs.version}</version>
+      <version>${checkstyle.version}</version>
     </dependency>
     <!-- required for Jsr305AnnotationsCheck -->
     <dependency>
@@ -239,7 +239,7 @@
           <dependency>
             <groupId>com.puppycrawl.tools</groupId>
             <artifactId>checkstyle</artifactId>
-            <version>${checkstyle.eclipse-cs.version}</version>
+            <version>${checkstyle.version}</version>
           </dependency>
         </dependencies>
         <executions>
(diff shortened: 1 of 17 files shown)


sevntu-checks/pom.xml
```diff
@@ -16,8 +16,8 @@
   <properties>
     <project.build.sourceEncoding>iso-8859-1</project.build.sourceEncoding>
     <!-- It is compile dependency to checkstyle, this version has to be the same as eclipse-cs depends on -->
-    <checkstyle.eclipse-cs.version>8.18</checkstyle.eclipse-cs.version>
-    <checkstyle.configLocation>https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-${checkstyle.eclipse-cs.version}/config/checkstyle_checks.xml</checkstyle.configLocation>
+    <checkstyle.version>8.23</checkstyle.version>
```
rnveach: This isn't possible.
Sevntu-checks may be fine, but eclipse-sevntu won't work because eclipse-cs hasn't been upgraded yet.
Sevntu is currently entirely dependent on eclipse-cs version. New property name is very misleading.

I don't see how this can be merged unless we split apart version numbering between sevntu and eclipse-sevntu.

pzygielo: I still can get the meaning of `checkstyle.eclipse-cs.version` then.
How it is connected with eclipse-cs, given
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/9765a3bf0f02272f921aec264667328f2ca60339/.ci/travis.sh#L15-L18 
and https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/9765a3bf0f02272f921aec264667328f2ca60339/eclipsecs-sevntu-plugin/pom.xml#L27-L31

Is that rather CS - eclipse-cs - sevntu compatibility level or something?
How does `checkstyle.eclipse-cs.version` influence dependency to eclipse-cs?

rnveach: Sevntu is a library, just like checkstyle. Sevntu requires the checkstyle dependency.

eclipsecs-sevntu-plugin is a plugin which has sevntu as a dependency.
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/eclipsecs-sevntu-plugin/pom.xml#L34-L36
It also has a dependency on the main eclipse-cs project which requires it's own checkstyle version.
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/eclipsecs-sevntu-plugin/pom.xml#L38-L47

This is the great circle of life.

So eclipsecs-sevntu-plugin for version X requires sevntu version X which then relies on checkstyle version Y through sevntu and checkstyle version Y through eclipse-cs.

If you upgrade checkstyle version to Z, sevntu pulls that in and may work fine by itself. The problem is then eclipsecs-sevntu-plugin pulls in that same sevntu version and tries to work with that version of checkstyle while also working with eclipse-cs version of checkstyle.
Since 2 versions aren't compatible by the updates you are required to make, something will break if we move sevntu's checkstyle version to be later than what eclipse-cs has released.

The only way to really fix this issue is either have eclipse-cs update to newer versions or break our current versioning system and allow eclipsecs-sevntu-plugin to stay stale until eclipse-cs is updated.

I hope that helps explain the problem better.

rnveach: https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/9765a3bf0f02272f921aec264667328f2ca60339/.ci/travis.sh#L15-L18 
Since sevntu is on checkstyle 8.18 , it looks like this SH file is outdated by displaying 8.12

pzygielo: Thank you for your rich response.

>  allow eclipsecs-sevntu-plugin to stay stale until eclipse-cs is updated and allow new releases of sevntu alone.

[As I do not use that plugin...] I'd second such solution! :grin: 

> I hope that helps explain the problem better.

Indeed it does, thank you.

pzygielo:
> Since 2 versions aren't compatible by the updates you are required to make in this PR, something will break if we move sevntu's checkstyle version to be later than what eclipse-cs has released.

Understood.
Thank you for review and big picture of dependencies.

sevntu-checks/pom.xml
rnveach: @romani ping

pzygielo: > allow eclipsecs-sevntu-plugin to stay stale until eclipse-cs is updated and allow new releases of sevntu alone.

Yesterday I was enthusiastic about that idea as it would solve _my_ problem. But now that I put some boxes and arrows on paper I really like that. (Not for plugin to be stale, but to separate them.) And perhaps sonar plugin as well...

Oh, I just found #705, I'll get familiar with that. It has no labels nor milestones set though. I assume you are processing it in background anyway.

rnveach: I put an approved label on it. We don't put on milestones until it is completed.

PR #775, opened 2019-09-22
config: upgrade maven-checkstyle-plugin to 3.1.0

rnveach:


rnveach:
eclipse analysis is out of date and that is why travis failed.

PR #779, opened 2019-09-29
config: ensure files in all commits are examined by CI

rnveach:
Starting to sink CIs between checkstyle and sevntu


romani:
@rnveach , should we close this ?
as this does not work in main repo ... so probably sevntu do not need this for now.
ones it become stable in main repo, we can copy it to sevntu.

Issue #782, opened 2019-11-19
Upgrade checkstyle to 8.26


romani:
combines #776 and #745

romani:
fix is merged

PR #783, opened 2019-11-19
Issue #782: upgrade to checksyle and eclipse-cs 8.26

romani:
Issue #782


romani:
merged manually  ....
problem with log exceeding is addressed in separate commit.

Issue #785, opened 2019-12-06
execute latest checkstyle version during verify phase


romani:
build problems in checkstyle repo:
```
✔ ~/java/github/romani/checkstyle [infra-activate-sevntu-checks L|✚ 1] 
$ ./.ci/wercker.sh no-error-sevntu-checks
CS_version: 8.28-SNAPSHOT
Target project sevntu.checkstyle is already cloned, latest changes will be fetched
[INFO] Error stacktraces are turned on.
[INFO] Scanning for projects...
[INFO] Inspecting build with total of 1 modules...
[INFO] Installing Nexus Staging features:
[INFO]   ... total of 1 executions of maven-deploy-plugin replaced with nexus-staging-maven-plugin
[INFO] 
[INFO] -------------< com.github.sevntu-checkstyle:sevntu-checks >-------------
[INFO] Building sevntu-checks 1.36.0
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-enforcer-plugin:1.2:enforce (enforce-maven) @ sevntu-checks ---
[INFO] 
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ sevntu-checks ---
[INFO] Using 'iso-8859-1' encoding to copy filtered resources.
[INFO] Copying 6 resources
[INFO] 
[INFO] --- maven-compiler-plugin:3.3:compile (default-compile) @ sevntu-checks ---
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 66 source files to /home/rivanov/java/github/romani/checkstyle/.ci-temp/sevntu.checkstyle/sevntu-checks/target/classes
[INFO] 
[INFO] --- maven-checkstyle-plugin:3.1.0:check (default) @ sevntu-checks ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  3.820 s
[INFO] Finished at: 2019-12-06T09:17:43-08:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:
3.1.0:check (default) on project sevntu-checks: Failed during checkstyle configuration:
 cannot initialize module TreeWalker - cannot initialize module JavadocContentLocation
 - Unable to instantiate 'JavadocContentLocation' class, it is also not possible to 
[...]

PR #786, opened 2019-12-06
Issue #785: use latest checkstyle version in verify phase

romani:
issue #785


romani:
now we have two different versions for compile and for verify phases.
Looks like there is no conflict.

Issue #789, opened 2020-01-01
Support Junit5 in RequireFailForTryCatchInJunit


pbludov:
Check [RequireFailForTryCatchInJunit](http://sevntu-checkstyle.github.io/sevntu.checkstyle/apidocs/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.html)

$ cat TestClass.java
```
import org.junit.jupiter.api.Test; // Not org.junit.Test

class InputRequireFailForTryCatchInJunit5Check {
    @Test
    public void method() {
        try {
            System.exit(0);
        }
        catch (Exception ignored) {
        }
    }
}
```
$ cat TestConfig.xml
```
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
          "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
<module name="Checker">
  <module name="TreeWalker">
    <module name="RequireFailForTryCatchInJunit"/>
  </module>
</module>
```
$java -classpath sevntu-checks-1.36.0.jar:checkstyle-8.28-all.jar com.puppycrawl.tools.checkstyle.Main -c TestConfig.xml TestClass.java`
```
Starting audit...
Audit done.
```
Expected violation:
```
[ERROR] TestClass.java:6:9: try/catch either requires a fail at the end of the try clause for junit tests or to remove the try/catch completely. [RequireFailForTryCatchInJunit]
```

Junit 5 API provides new `@Test` annotation in package `org.junit.jupiter.api`. The check should be aware of this annotation.

romani:
fix is merged.

PR #790, opened 2020-01-01
Issue #789: Support Junit5 in RequireFailForTryCatchInJunit

pbludov:
Issue #789

Support for Junit5 tests

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
@@ -19,6 +19,9 @@
 
 package com.github.sevntu.checkstyle.checks.coding;
 
+import java.util.Arrays;
+import java.util.List;
+
 import com.github.sevntu.checkstyle.SevntuUtil;
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
@@ -77,7 +80,15 @@ public class RequireFailForTryCatchInJunitCheck extends AbstractCheck {
     /**
      * Fully qualified junit test annotation.
      */
-    private static final String FQ_JUNIT_TEST = "org.junit.Test";
+    private static final List<String> FQ_JUNIT_TESTS = Arrays.asList(
+            "org.junit.Test",
+            "org.junit.jupiter.api.Test");
+    /**
+     * Fully qualified junit `fail` methods.
+     */
+    private static final List<String> FQ_JUNIT_FAIL_METHODS = Arrays.asList(
+            "org.junit.Assert.fail",
+            "org.junit.jupiter.api.Assertions.fail");
     /**
      * JUnit's fail assertion method name.
      */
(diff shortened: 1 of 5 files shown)


sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
```diff
@@ -88,9 +99,13 @@
...
     /**
-     * {@code true} if the junit assert is imported.
+     * {@code true} if the junit 4 assert is imported.
+     */
+    private boolean importJunit4Assert;
+    /**
+     * {@code true} if the junit 5 assertions is imported.
      */
-    private boolean importAssert;
+    private boolean importJunit5Assertions;
```
rnveach: Why do we need to identify junit 4 support versus 5? If someone imports an assertion, then it is slightly hard to use another assert without using the fully qualified class.

Most common cases people will use one junit version or the other. I did happen to mix and match 4 and 5 in main repo since my eclipse doesn't support junit 5 and I was able to run junit 4 Test with junit 5 assertions.

pbludov: We can clone this check to verify Junit5 assertions only. Users will have to specify which check they want to use in their config. @romani what do you think?

rnveach: > We can clone this check to verify Junit5 assertions only. Users will have to specify which check they want to use in their config.

This doesn't specify why it is needed. Why do we need to know exact junit version user is using for validation?

pbludov: > Why do we need to know exact junit version user is using for validation?

We do not need it. We need to know which assertion class is imported: `Assert` from Junit 4 or `Assertions` from Junit 5. To avoid false positives like this:
```
import org.junit.Assert;
import org.example.Assertions;

// ....
    try {
       foo();
       Assertions.fail("I'm a wrong fail method"); 
    }
```
For static import of the method `fail` and for annotation `Test` we do not need to know which framework is in use.

rnveach: Thank you, I understand now. I am fine to keep it all in 1.

rnveach:
Regression: http://rveach.no-ip.org/checkstyle/regression/reports/286/

Only violations are in spring-integration on the end. They look valid to me.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/RequireFailForTryCatchInJunitCheck.java
romani: @pbludov , @rnveach , ideally it would be better to make it configurable from user properties to user himself define classes to consider as assert/fail. Junit3/4 was with us too long, but there bunch of other tools and custom `fail` methods, and this Check only care that some recognizable method is called.

Issue #793, opened 2020-01-12
Remove usages of `DetailAST::getNumberOfChildren`


pbludov:
The interface `DetailAST` currently contains two methods to return the count of children:

* [getChildCount](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/DetailAST.java#L36)
* [getNumberOfChildren](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/DetailAST.java#L119)

Both methods are implemented in the same way (`getChildCount` has an optimization):

https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/DetailAstImpl.java#L166

https://github.com/Distrotech/antlr2/blob/master/antlr/BaseAST.java#L67

We should drop one of them in the future. In the meantime, let's update the code to use only the optimized version.

rnveach:
> getNumberOfChildren
> BaseAST.java

I agree if they are the same and this method is coming from ANTLR. It will most likely go away if we ever upgrade to newer antlr and would have to re-implement it.

rnveach:
@pbludov Do you plan to make same issue in main library?

pbludov:
> @pbludov Do you plan to make same issue in main library?

Yes, this method should be deprecated in the main library. And later removed.

romani:
@pbludov , is this issue done ?

pbludov:
> @pbludov , is this issue done ?

This issue is done, `node.getFirstChild() != null` should be replaced with `node.hasChildren()` after next release of Checkstyle.

PR #794, opened 2020-01-12
Remove usages of DetailAST::getNumberOfChildren

pbludov:
Issue #793

Usages of `node.getNumberOfChildren() > 0` replaced with `node.getFirstChild() != null`.
This is less readable, may be there should be method
```java
default boolean hasChildren() {
    return firstChild != null;
}
```
in the DetailAst interface?


rnveach:
@romani CI seems broken as travis says everything is skipped. I noticed this because commit is not correctly formatted.

pbludov:
> commit is not correctly formatted.

done

romani:
Auto Disablement by commits should be disabled https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/.travis.yml#L81 it is too unreliable.

PR #804, opened 2020-03-05
config: upgrade maven-checkstyle-plugin to 3.1.1

nmancus1:
Checkstyle is looking to remove some deprecated methods and CI noticed that you are not on the latest maven-checkstyle-plugin which is still using some deprecated methods. 

Issue: checkstyle/checkstyle/issues/7190
PR: checkstyle/checkstyle#7778


rnveach:
thanks @nmancus1

PR #805, opened 2020-03-05
Update AvoidHidingCauseExceptionCheck.java

yajna-pandith:
Issue #7476
While I have been able to reduce the complexity of the code,  I am not very satisfied with the verbosity introduced by checkstyle's unnecessary braces and sonar's complaint. I will post the PR.  I would like you to look into  sonar errors with respect to unnecessary braces in lambda and suggest me if I should tinker checkstyle configuration for the same


rnveach:
@yajna-pandith ping
Do you plan to continue this? Also can you attach an issue link to this. I can't find what this is about.

Issue #819, opened 2020-06-20
Use Jacoco instead of Cobertura for coverage check


pbludov:
This is one of prerequisites to build sevntu.checkstyle with Java 11. Same as https://github.com/checkstyle/checkstyle/issues/5586

Stopper for https://github.com/checkstyle/checkstyle/pull/8326

romani:
fix is merged

PR #820, opened 2020-06-20
Issue #819: replace cobertura with jacoco for coverage

pbludov:
Issue #819 

Local build with Java 14 succeeded on my PC.


romani:
Travis failire:
```
Caused by: java.lang.IllegalStateException: Class com/github/sevntu/checkstyle/SevntuUtil is already instrumented.
    at org.jacoco.core.internal.instr.InstrSupport.assertNotInstrumented (InstrSupport.java:89)
    at org.jacoco.core.internal.analysis.ClassAnalyzer.visitField (ClassAnalyzer.java:106)
```

pbludov:
> Caused by: java.lang.IllegalStateException: Class com/github/sevntu/checkstyle/SevntuUtil is already instrumented.

This can be fixed by removing PowerMock from Sevntu.Checkstyle. It is used in one test only.

Issue #822, opened 2020-08-12
Introduction of YML metadata for eclipse-cs integration


gaurabdg:
It was decided by @romani and me, that all eclipse-cs extension projects will have `eclipse-metadata.yml` file which would consist metadata required during automatically generating check metadata and integrating with eclipse-cs.

For now, the required eclipse cs configuration metadata are the `RuleGroupMetadata.name` and related data.

kaushal-2002:
Hi, I am a beginner at this issue can you help me? How to work on it?

rnveach:
I am marking this issue as blocked because of issues identified at https://github.com/checkstyle/eclipse-cs/pull/317#issuecomment-910812549 .

eclipse-cs metadata is not working with this YML in it. I don't know if it is because of the combination of XML/YML (old/new) or something else.

In order to restore functionality of sevntu's eclipse plugin while investigating eclipse-cs more, the commit adding the YML should be reverted.

Even if eclipse-cs' functionality is fixed, I feel this should be blocked by https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/864 and wait until we convert everything to the new format.

PR #823, opened 2020-08-12
Issue #822: Introduction of YML metadata for eclipse-cs integration

gaurabdg:
#822 
Details are mentioned in the issue

### DIFF
--- a/eclipsecs-sevntu-plugin/src/com/github/sevntu/eclipse-metadata.yml
+++ b/eclipsecs-sevntu-plugin/src/com/github/sevntu/eclipse-metadata.yml
@@ -1,0 +1,6 @@
+---
+ruleGroups:
+  - name: "SevNTU checks"
+    description: "Check extensions developed by SevNTU"
+    priority: 50
+    package: "com.github.sevntu"
--- a/eclipsecs-sevntu-plugin/pom.xml
+++ b/eclipsecs-sevntu-plugin/pom.xml
@@ -66,6 +66,7 @@
               <dir>src</dir>
               <excludes>
                 <exclude>**/*.properties</exclude>
+                <exclude>**/*.yml</exclude>
               </excludes>
             </validationSet>
           </validationSets>


eclipsecs-sevntu-plugin/src/com/github/sevntu/eclipse-metadata.yml
```diff
@@ -0,0 +1,4 @@
+---
+extensionName: "com.github.sevntu"
+ruleGroupName: "SevNTU checks"
+ruleGroupDesc: "Check extensions developed by SevNTU"
```
romani, line 4: In original config there is some number also  that might be used to define position of extension in tree of modules in eclipse-cs UI

romani: Where we can define `priority`
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/83377c1c1d25a95f36837d4b3fa6e721d108a8aa/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/annotation/checkstyle-metadata.xml#L6

?

gaurabdg: @romani  I will add a field for it.
But why is `priority` different in some packages whereas rule group is same?
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/83377c1c1d25a95f36837d4b3fa6e721d108a8aa/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/coding/checkstyle-metadata.xml#L6

romani: @Gaurav-Punjabi , it is mistake, it should be the same.

gaurabdg: > Where we can define `priority`
> https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/83377c1c1d25a95f36837d4b3fa6e721d108a8aa/eclipsecs-sevntu-plugin/src/com/github/sevntu/checkstyle/checks/annotation/checkstyle-metadata.xml#L6
> 
> ?

@romani Done.

romani: We should give ability to use multiple groups from single extension in yml, please do

gaurabdg: I didnt understand why would we need multiple groups, please provide and example/use case

gaurabdg: > We should give ability to use multiple groups from single extension in yml, please do

@romani Done. The explanation is present in the issue description

gaurabdg:
> ```
> ruleGroups:
>   - 
>     name: "SevNTU checks"
>     description: "Check extensions developed by SevNTU"
>     priority: 50
>     package: "com.github.sevntu"
> ```

Done

PR #838, opened 2021-04-02
config: bump checkstyle version to 8.37 (same as elipse-cs version)

romani:


rnveach:
@romani Shouldn't all versions be kept in synch?

https://github.com/sevntu-checkstyle/sevntu.checkstyle/search?l=Maven+POM&q=%228.29%22

romani:
Not all, some depends on eclipse cs , others are maven build run time

rnveach:
@romani But then how can sonar work when it is using an old version and actual sevntu and others is using a newer version? What about breaking changes between the 2?

romani:
Eclipse extension is always behind.
All extensions are independent, unfortunately bond by same version of project

PR #845, opened 2021-05-20
minor: update checks (related to checkstyle #3145)

nmancus1:
Related to https://github.com/checkstyle/checkstyle/pull/9922,  before we update sevntu check's `checkstyle.eclipse-cs.version` to 8.43, we will need to merge this PR to avoid regression in `AvoidNotShortCircuitOperatorsForBooleanCheck` and `NoNullForCollectionReturnCheck`.

Diff report from checkstyle#9922: https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/24f673f_2021192452/reports/diff/index.html (regression)

Diff report after this PR: https://nmancus1.github.io/sevntu-diff-3145-b/index.html (no regression)

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
@@ -21,9 +21,11 @@ package com.github.sevntu.checkstyle.checks.coding;
 
 import java.util.LinkedList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import com.puppycrawl.tools.checkstyle.api.AbstractCheck;
 import com.puppycrawl.tools.checkstyle.api.DetailAST;
+import com.puppycrawl.tools.checkstyle.api.FullIdent;
 import com.puppycrawl.tools.checkstyle.api.TokenTypes;
 import com.puppycrawl.tools.checkstyle.utils.CheckUtil;
 
@@ -80,9 +82,9 @@ public class AvoidNotShortCircuitOperatorsForBooleanCheck extends AbstractCheck 
     public static final String MSG_KEY = "avoid.not.short.circuit.operators.for.boolean";
 
     /**
-     * A "boolean" String.
-     * */
-    private static final String BOOLEAN = "boolean";
+     * Pattern to match boolean types, including array types.
+     */
+    private static final Pattern BOOLEAN_TYPE_PATTERN = Pattern.compile("^boolean(\\[[^]]*])*");
 
     /**
      * A list contains all names of operands, which are used in the current
@@ -140,8 +142,11 @@ public class AvoidNotShortCircuitOperatorsForBooleanCheck extends AbstractCheck 
      * @return "true" if current method or variable has a Boolean type.
      */
     private static boolean isBooleanType(final DetailAST node) {
-        return BOOLEAN.equals(CheckUtil.createFullType(
-                node.findFirstToken(TokenTypes.TYPE)).getText());
+        final FullIdent methodOrVariableType =
+                CheckUtil.createFullType(node.findFirstToken(TokenTypes.TYPE));
+        return BOOLEAN_TYPE_PATTERN
+                .matcher(methodOrVariableType.getText())
+                .find();
     }
 
     /**
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/NoNullForCollectionReturnCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/NoNullForCollectionReturnCheck.java
@@ -181,10 +181,11 @@ public class NoNullForCollectionReturnCheck extends AbstractCheck {
      * @return true, when method return collection.
      */
     private boolean isReturnCollection(DetailAST methodDef) {
-        DetailAST methodType = methodDef.findFirstToken(TokenTypes.TYPE);
-        methodType = methodType.getFirstChild();
-        return methodType.getType() == TokenTypes.ARRAY_DECLARATOR
-                || collectionList.contains(methodType.getText());
+        final DetailAST methodType = methodDef.findFirstToken(TokenTypes.TYPE);
+        final boolean isArrayType =
+                methodType.findFirstToken(TokenTypes.ARRAY_DECLARATOR) != null;
+        return isArrayType
+                || collectionList.contains(methodType.getFirstChild().getText());
     }
 
     /**
--- a/sevntu-checks/src/test/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheckTest.java
+++ b/sevntu-checks/src/test/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheckTest.java
@@ -78,4 +78,16 @@ public class AvoidNotShortCircuitOperatorsForBooleanCheckTest extends AbstractMo
             CommonUtil.EMPTY_STRING_ARRAY);
     }
 
+    @Test
+    public final void testBitwiseOrAfterArrayIndex() throws Exception {
+        final DefaultConfiguration checkConfig =
+                createModuleConfig(AvoidNotShortCircuitOperatorsForBooleanCheck.class);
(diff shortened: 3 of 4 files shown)


sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
```diff
@@ -140,8 +141,11 @@ public final void visitToken(final DetailAST detailAST) {
      * @return "true" if current method or variable has a Boolean type.
      */
     private static boolean isBooleanType(final DetailAST node) {
-        return BOOLEAN.equals(CheckUtil.createFullType(
-                node.findFirstToken(TokenTypes.TYPE)).getText());
+        final FullIdent methodReturnType =
+                CheckUtil.createFullType(node.findFirstToken(TokenTypes.TYPE));
+        return methodReturnType
+                .getText()
+                .contains(BOOLEAN);
```
rnveach, line 148: Original code was `equals` and now this changed to `contains`. Doesn't this mean we will get false positives with things like `AtomicBoolean`?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/NoNullForCollectionReturnCheck.java
```diff
@@ -181,10 +181,9 @@ public void visitToken(DetailAST detailAST) {
      * @return true, when method return collection.
      */
     private boolean isReturnCollection(DetailAST methodDef) {
-        DetailAST methodType = methodDef.findFirstToken(TokenTypes.TYPE);
-        methodType = methodType.getFirstChild();
-        return methodType.getType() == TokenTypes.ARRAY_DECLARATOR
-                || collectionList.contains(methodType.getText());
+        final DetailAST methodType = methodDef.findFirstToken(TokenTypes.TYPE);
+        final boolean isArrayType = methodType.findFirstToken(TokenTypes.ARRAY_DECLARATOR) != null;
```
rnveach, line 185: Is this safe for current sevntu's version of checkstyle or must this rely solely on newest checkstyle?

See https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/pom.xml#L18-L19

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
nmancus1: Not `AtomicBoolean`, no. It would only match strings that contain lowercase `boolean`.  You can't use logical operators on `AtomicBoolean` , anyway. This would match a return type of a class that had lowercase `boolean` in it though; but the code wouldn't compile if we used a bitwise operator on anything besides a boolean or an int anyway. So, we could have a false positive on an int that was named `myboolean`. Previously, `createFullType()` was not returning the `[]` reliably for some 1-D arrays ( same bug from the main PR), so it was a happy accident that this worked in the first place.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/NoNullForCollectionReturnCheck.java
nmancus1: This relies on 8.43 (or whenever main PR is merged)

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
nmancus1: I can just use a matcher here, to prevent false positives.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/NoNullForCollectionReturnCheck.java
nmancus1: Actually, we should be good to merge this now, let's see what CI says.

rnveach: I would think regression might show more than CI.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
rnveach: Ok, `contains` is case sensitive. But then makes it a concern if a user creates a class `Funwithboolean`.

nmancus1: But, you would not be able to return the result of a bitwise operator on, or with any class except `int` and `boolean`. It would not compile.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/NoNullForCollectionReturnCheck.java
nmancus1: https://nmancus1.github.io/sevntu-diff-3145-b/index.html

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
rnveach: Yes, but it is not clear from this snippet of code as that is in other parts of the class. This method just says `isBoolean` and we don't know who might make use of it in the future.

Then I just have 1 final question since you changed this to an expression. Why not add `^$` to enforce you are only looking at exactly a `boolean` (and any `[]`s)?

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/NoNullForCollectionReturnCheck.java
rnveach: I am not following what the new URL is. Both list the same "master" and "patch" commit.

nmancus1: The report is using sevntu checks built from this PR.

rnveach: Then how is it different from the one in the first post?

nmancus1: The one in the first post is using sevntu checks from the master branch, ran using report generation in https://github.com/checkstyle/checkstyle/pull/9922.

So, this report is ran using sevntu checks built form master, which shows the regression:  https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/24f673f_2021192452/reports/diff/index.html

And this report was generated with the same Checkstyle codebase, but with sevntu checks built from the code in this PR: https://nmancus1.github.io/sevntu-diff-3145-b/index.html , showing that there is no regression once this PR is merged.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidNotShortCircuitOperatorsForBooleanCheck.java
nmancus1: done

romani:
why you think it is not a regression 
https://checkstyle-diff-reports.s3.us-east-2.amazonaws.com/24f673f_2021192452/reports/diff/android-launcher/index.html#A1 ?

http://sevntu-checkstyle.github.io/sevntu.checkstyle/apidocs/com/github/sevntu/checkstyle/checks/coding/NoNullForCollectionReturnCheck.html

>  Returning Null is error prone as developers forget to review method implementation or Java doc for nuances of Null meaning.

romani (approved):

PR #859, opened 2021-08-11
config: bump eclipse-cs to 8.44.0-SNAPSHOT

romani:
requirements for breaking activity in checkstyle before 9.0


rnveach (approved):
Sevntu-checks was already bumped. We should add a comment to note both should be the same to both areas.

Issue #857, opened 2021-07-27
Eclipse update site wrong 


mbert:
According to the documentation this URL should work:  https://sevntu-checkstyle.github.com/sevntu.checkstyle/update-site/
However it does not. When configuring it as an update site for Eclipse, it emits this error message:
```
No repository found at https://sevntu-checkstyle.github.com/sevntu.checkstyle/update-site/.
```
Tested with Eclipse Version: 2021-06 (4.20.0)

So either the update site is broken or the documentation is wrong.

romani:
@rnveach , can you recheck in your eclipse ?

rnveach:
See https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/856#issuecomment-877734372 which details the correct site and the fix we made.

@mbert Where are you saying our documentation is still wrong?

@romani I recommend adding an index page to our GH page in this folder to atleast say they hit the right location in their browser, and that they should give this location to Eclipse.

rnveach:
@romani Scanning our code base, `http://sevntu-checkstyle.github.com/` seems like a wrong URL. It or `http://sevntu-checkstyle.github.com/sevntu.checkstyle` does not work, so we should change them.

I am seeing it being used multiple times in https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/eclipsecs-sevntu-plugin-feature/feature.xml .

romani:
@mbert , can you send PR to fix wrong links

nkavian:
Having `.com` instead of `.io` on the [Distribution page](https://sevntu-checkstyle.github.io/sevntu.checkstyle/#distribution) caused me to lose many hours of work.

![image](https://user-images.githubusercontent.com/1127421/130332277-f06290cc-958e-4261-b106-b5380afaa65a.png)

:(

nkavian:
Although 1.40 SevNTU installs into Eclipse, the configuration menu is still empty. This is all very frustrating!

![image](https://user-images.githubusercontent.com/1127421/130332395-aac9442d-1288-49fe-aba8-cf52c550e33c.png)

romani:
>caused me to lose many hours of work.

Sorry about this, please send Pull Request to fix it.

rnveach:
Continuing on with https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/857#issuecomment-906881832 ,

[...]

PR #861, opened 2021-08-21
minor: fix broken URLs

nkavian:
Fix the URLs that should have used `.io` instead of `.com`.

https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/857#issuecomment-903161782


romani (approved):
Thanks a lot for update

romani:
I can release changes as we get at least one functional update.

PR #863, opened 2021-08-23
Issue #857: added main index to update site

rnveach:
Updated gh-pages site so there is a main page for index and not be confused with the wrong URL.

I think CI can be ignored since there is no setup for it on gh-pages.


romani (approved):
Let's try

romani:
@rnveach , http://sevntu-checkstyle.github.io/sevntu.checkstyle/update-site is still 404

PR #865, opened 2021-09-05
config: switch travis over to the partner queue solution

rnveach:
Travis issue is being resolved. Testing out switching to new system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,9 @@
+os: linux
+arch:
+  - ppc64le
+# Above is required as is for Partner Queue Solution.
+# DO NOT MODIFY!
+
 language: java
 sudo: false
 


pbludov:
org.apache.maven.InternalErrorException: Internal error: java.lang.RuntimeException: Could not determine SWT implementation fragment bundle for environment {osgi.os=linux, osgi.ws=gtk, org.eclipse.update.install.features=true, osgi.arch=ppc64le}

It looks like OpenJDK8 has only partial support  for ppc64le. Maybe this can be solved with some modern version of JDK?

romani:
Swt issue is resolved by using s390x 
See https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/867

PR #867, opened 2021-09-18
infra: use free arch in Travis to avoid usage of credits

romani:
Travis allow unlimited execution only on special arch
https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z

s390x: no jdk11 support, only jre (problmes with jacoco, javadoc), jdk8 support is problematic.
Question to support: https://travis-ci.community/t/arch-s390x-has-jre-but-most-builds-of-java-needs-jdk/12143/2

### DIFF
--- a/.ci/travis.sh
+++ b/.ci/travis.sh
@@ -56,9 +56,10 @@ sevntu-checks)
   mvn -e -Pcoverall install
   mvn -e verify -Pno-validations,selftesting
   mvn -e javadoc:javadoc
-  if [[ $TRAVIS == 'true' ]]; then
-    mvn -e -Pcoverall jacoco:report coveralls:report
-  fi
+  # until problem coverall back to online
+  #if [[ $TRAVIS == 'true' ]]; then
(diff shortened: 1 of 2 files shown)


.ci/travis.sh
```diff
@@ -56,9 +56,10 @@ sevntu-checks)
   mvn -e -Pcoverall install
   mvn -e verify -Pno-validations,selftesting
   mvn -e javadoc:javadoc
-  if [[ $TRAVIS == 'true' ]]; then
-    mvn -e -Pcoverall jacoco:report coveralls:report
-  fi
+  # until problem coverall back to online
```
rnveach, line 59: How will we know it's back online?

rnveach (approved):

.ci/travis.sh
romani: We do not use this service, it has fancy badge and html report, I do not remember when last time I used it.

PR #872, opened 2021-09-30
minor: Fix usage of tags inside description field to fix parsing problem with Sonarqube 9.x

KTannenberg:
see details below in comments


rnveach:
@KTannenberg Please provide a screenshot of the old and new descriptions in sonar. Are these descriptions suppose to be pure text or is the HTML allowed and displayed correctly?

Issue #883, opened 2022-02-20
upgrade minimal required jdk version to 11 and eclipse-cs 10.0.0


pbludov:
To match https://github.com/checkstyle/checkstyle/issues/9146 this project should be upgraded too.

romani:
fix is merged

rnveach:
@romani @nrmancuso @pbludov Is sevntu really JDK 11?

Eclipse is still importing this as JDK 8.
Probably because we still have 1.8 references like https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/pom.xml#L166 .
Our CI eclipse compiler still says 1.8. https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/.ci/eclipse-compiler-javac.sh#L42

We look incomplete here.

romani:
ok,  better to remove all references to 1.8

romani:
Merged

PR #884, opened 2022-02-20
Issue #883: upgrade minimal required jdk version to 11

pbludov:
Issue #883 

Some Guava code replaced with JDK alternatives.

### DIFF
--- a/sevntu-checks/pom.xml
+++ b/sevntu-checks/pom.xml
@@ -16,7 +16,7 @@
   <properties>
     <project.build.sourceEncoding>iso-8859-1</project.build.sourceEncoding>
     <!-- It is compile dependency to checkstyle, this version has to be the same as eclipse-cs depends on -->
-    <checkstyle.eclipse-cs.version>9.1</checkstyle.eclipse-cs.version>
+    <checkstyle.eclipse-cs.version>10.0</checkstyle.eclipse-cs.version>
     <!-- verify time version -->
     <checkstyle.version>10.2</checkstyle.version>
     <checkstyle.configLocation>https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-${checkstyle.version}/config/checkstyle_checks.xml</checkstyle.configLocation>
--- a/sevntu-checks/pom.xml
+++ b/sevntu-checks/pom.xml
@@ -25,7 +25,7 @@
     <maven.jacoco.plugin.version>0.8.7</maven.jacoco.plugin.version>
     <maven.pmd.plugin.version>3.11.0</maven.pmd.plugin.version>
     <pmd.version>6.38.0</pmd.version>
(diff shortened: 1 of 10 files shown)


sevntu-checks/pom.xml
```diff
@@ -25,7 +25,7 @@
     <maven.jacoco.plugin.version>0.8.7</maven.jacoco.plugin.version>
     <maven.pmd.plugin.version>3.11.0</maven.pmd.plugin.version>
     <pmd.version>6.38.0</pmd.version>
-    <java.version>1.8</java.version>
+    <java.version>11</java.version>
```
rnveach, line 28: When this is ready for merge, can you provide an example of using the Checkstyle JDK11 All JAR with Sevntu's JDK11 JAR and show a custom checks in Sevntu will work from the CLI using JDK11. I want  to verify an integration test to show sevntu works and we are not fooling ourselves with anything in our testing area.

I think manually testing will be enough, but I am not against from building a CI item to show this.

romani:
I think we need to bind this update to bump of plugins to 10.x version.

rnveach:
@pbludov it seems to me we can continue with this if eclipse-cs' bump is in maven.

rnveach:
Eclipse-cs now upgraded.
https://github.com/checkstyle/eclipse-cs/releases/tag/10.0.0

romani:
We need to move all jobs from wercker to Travis.

sevntu-checks/pom.xml
romani:
```
rivanov@p5510:/var/tmp$ cat Test.java 
public class Test
{
    private final                                                                                                     MyThing thing;

}
rivanov@p5510:/var/tmp$ cat config.xml 
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="LineLengthExtended"/>
    </module>
</module>
rivanov@p5510:/var/tmp$ java -classpath sevntu-checks-1.41.0.jar:checkstyle-10.2-all.jar com.puppycrawl.tools.checkstyle.Main -c config.xml Test.java 
Starting audit...
[ERROR] /var/tmp/Test.java:3: Line is longer than 80 characters. Current line length is 132. [LineLengthExtended]
Audit done.
Checkstyle ends with 1 errors.
```

sevntu version looks old, but it is how it working now (no snapshot, new version appears during release, but I took this jar from target folder)

romani (approved):
ok to merge

Wercker could be ignored, I am not sure why build is failing, but we need to move all jobs out of wercker.

Travis eclipsecs build is stuck in queue, I tested it on local.

PR #612, opened 2017-10-03
Issue #525: removed boolean complexity suppression and fixed violations

rnveach:
Issue #525

PR #886, opened 2022-04-11
add configuration option to allow nullable annotation on return

vdaniloff:
The great Jsr305 check resolves almost everyone needs for validation, but it does not allow nullable as return value annotation. 
On the other hand this is in fact commonly use practice. Even #612 uses that in code snippet.

A new parameter has been added to the validation to allow using nullable.


romani:
Thanks a lot, can you explain problem in details and example in issue.
Please create issue by format we need https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/.github/ISSUE_TEMPLATE.md
We need to approve issue as confirmation that you idea is correct.

Please use link to issue in commit message.

vdaniloff:
Created #887 and mentioned in commit

rnveach:
@vdaniloff @romani Is there agreement on the issue?

PR #920, opened 2022-10-25
config: re-enable sevntu selftesting

rnveach:
This undoes https://github.com/sevntu-checkstyle/sevntu.checkstyle/commit/eb45480f255354be534ab77f643aecfcbe4c8b7b

https://app.travis-ci.com/github/sevntu-checkstyle/sevntu.checkstyle/jobs/586424111#L355
> `[WARNING] 'build.plugins.plugin.(groupId:artifactId)' must be unique but found duplicate declaration of plugin org.apache.maven.plugins:maven-antrun-plugin @ line 285, column 15`

Maven cannot have 2 instances of the same plugins in the build section. We needed this because the 2 instances have different dependencies. One had CS 10.3.4 and the other had CS 10.0 and Sevntu. If you don't separate them like this, then maven tries to randomly combine them and the dependencies. This causes failures if the 2 versions are NOT compatible which is what I was seeing in the main project.

I had a note to say why this has to be a separate profile in the POM.


romani (approved):

romani:
Ok to merge.
We can rethink it later on

rnveach:
https://app.travis-ci.com/github/sevntu-checkstyle/sevntu.checkstyle/jobs/586628511#L841
Normal run.

https://app.travis-ci.com/github/sevntu-checkstyle/sevntu.checkstyle/jobs/586628511#L942
Profile run.

No more warnings.

PR #941, opened 2022-11-28
config: upgrade eclipsecs-sevntu-plugin to 10.4

rnveach:

PR #942, opened 2022-11-28
infra: fails build if checkstyle versions don't match with eclipse-cs

rnveach:
After looking at https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/941#issuecomment-1329429741 and options, we cannot combine POMs because eclipse-cs is not in maven, this is why we always have to install it before hand.

Instead, this change will manually verify checkstyle versions match up between different POMs. Fixed the first one where the main issue was identified.

This should fail without the referenced PR merged and should pass afterwards.

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -15,12 +15,25 @@ eclipse-cs)
                            | sed "s/-SNAPSHOT//")
   cd ../
   cd sevntu-checks
+  SEVNTU_ECLIPSE_CS_VERSION=$(mvn -e -q -Dexec.executable='echo' \
+                   -Dexec.args='${checkstyle.eclipse-cs.version}' \
+                   --non-recursive org.codehaus.mojo:exec-maven-plugin:1.3.1:exec)
+  echo "sevntus eclipse cs version: "$SEVNTU_ECLIPSE_CS_VERSION
   mvn -B -e clean install -Pno-validations
   cd ..
   checkout_from "https://github.com/checkstyle/eclipse-cs.git"
   cd .ci-temp/eclipse-cs/
   echo "Eclipse-cs tag: "$ECLIPSECS_TAG_NAME
   git checkout $ECLIPSECS_TAG_NAME
+  ECLIPSE_CS_CS_VERSION=$(mvn -e -q -Dexec.executable='echo' \
+                   -Dexec.args='${maven.sevntu.checkstyle.plugin.checkstyle.version}' \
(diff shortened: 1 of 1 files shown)


rnveach:
https://app.travis-ci.com/github/sevntu-checkstyle/sevntu.checkstyle/jobs/589652909#L422
There is no message, but CI failed as expected where expected.

.ci/validation.sh
```diff
@@ -15,12 +15,25 @@ eclipse-cs)
...
+  echo "sevntus eclipse cs version: "$SEVNTU_ECLIPSE_CS_VERSION
   mvn -B -e clean install -Pno-validations
   cd ..
   checkout_from "https://github.com/checkstyle/eclipse-cs.git"
   cd .ci-temp/eclipse-cs/
   echo "Eclipse-cs tag: "$ECLIPSECS_TAG_NAME
   git checkout $ECLIPSECS_TAG_NAME
+  ECLIPSE_CS_CS_VERSION=$(mvn -e -q -Dexec.executable='echo' \
+                   -Dexec.args='${maven.sevntu.checkstyle.plugin.checkstyle.version}' \
+                   --non-recursive org.codehaus.mojo:exec-maven-plugin:1.3.1:exec)
```
rnveach, line 30: Command is failing.

````
++ mvn -e -q -Dexec.executable=echo '-Dexec.args=${maven.sevntu.checkstyle.plugin.checkstyle.version}' --non-recursive org.codehaus.mojo:exec-maven-plugin:1.3.1:exec
+ ECLIPSE_CS_CS_VERSION='[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.3.1:exec (default-cli) on project net.sf.eclipsecs.parent: Command execution failed.: Process exited with an error: 1 (Exit value: 1) -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.3.1:exec (default-cli) on project net.sf.eclipsecs.parent: Command execution failed.
````

rnveach: POM variable is only available on 10.3.4 . We can't use this to test the current issue.

rnveach: Command fails with 10.4.0 as well. I am not sure why.

rnveach:
This won't work anyways. I am trying to pull in the verify time Checkstyle. They embed their JAR into the project, so I would need to verify a file name or something.

PR #972, opened 2022-12-25
minor: add date to xdoc template

rnveach:
Adds date and removes extra blank line on end.


rnveach:
Example of new display:
example

romani (approved):
It not xdoc, it is just HTML template

PR #1002, opened 2023-03-25
Pull #1002: Suppress check for keeping spacing between workflow job steps

stoyanK7:
We are implementing a check to enforce workflow job steps are spaced out in https://github.com/checkstyle/checkstyle/pull/12900

It is better to suppress it for now and implement it when PR gets merged and 10.9.3 is out.


stoyanK7:
I will open an issue to implement this when this PR is merged.

romani (approved):

PR #960, opened 2022-12-09
minor: restore features formatting and added remaining suppressions

rnveach:
Issue identified during deployment test.

````
diff --git a/eclipsecs-sevntu-plugin-feature/feature.xml b/eclipsecs-sevntu-plugin-feature/feature.xml
index 82b90d8..e718a9f 100644
--- a/eclipsecs-sevntu-plugin-feature/feature.xml
+++ b/eclipsecs-sevntu-plugin-feature/feature.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<feature id="com.github.sevntu.checkstyle.checks.feature"
-   label="Extension for eclipse-cs plugin with additional Checks" version="1.44.0">
+<feature id="com.github.sevntu.checkstyle.checks.feature" label="Extension for eclipse-cs plugin with additional Checks" version="1.44.1">
 
    <description url="http://sevntu-checkstyle.github.io/sevntu.checkstyle/">
       http://sevntu-checkstyle.github.io/sevntu.checkstyle/
@@ -17,15 +16,11 @@
    </license>
 
    <url>
-      <discovery label="Sources on GitHub"
-         url="https://github.com/sevntu-checkstyle/sevntu.checkstyle"/>
-      <discovery label="SevNTU checks site"
-         url="http://sevntu-checkstyle.github.io/sevntu.checkstyle/"/>
-      <discovery label="Eclipse Update site"
-         url="http://sevntu-checkstyle.github.io/sevntu.checkstyle/update-site/"/>
+      <discovery label="Sources on GitHub" url="https://github.com/sevntu-checkstyle/sevntu.checkstyle"/>
+      <discovery label="SevNTU checks site" url="http://sevntu-checkstyle.github.io/sevntu.checkstyle/"/>
+      <discovery label="Eclipse Update site" url="http://sevntu-checkstyle.github.io/sevntu.checkstyle/update-site/"/>
    </url>
 
-   <plugin id="eclipsecs-sevntu-plugin" download-size="0" install-size="0" version="1.44.0"
-      unpack="false"/>
+   <plugin id="eclipsecs-sevntu-plugin" download-size="0" install-size="0" version="1.44.1" unpack="false"/>
 
 </feature>
````

PR #1005, opened 2023-04-12
Remove broken discovery sites

Bananeweizen:
Discovery sites are used by the p2 installation system to discover additional update sites containing eclipse plugins and features. Adding arbitrary websites as discovery sites is an error and will lead to the Eclipse update mechanism trying to interpret those websites as update sites, failing, and then adding error messages in the log on each update.

### DIFF
--- a/eclipsecs-sevntu-plugin-feature/feature.xml
+++ b/eclipsecs-sevntu-plugin-feature/feature.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<feature id="com.github.sevntu.checkstyle.checks.feature" label="Extension for eclipse-cs plugin with additional Checks" version="1.44.1">
+<feature
+      id="com.github.sevntu.checkstyle.checks.feature"
+      label="Extension for eclipse-cs plugin with additional Checks"
+      version="1.44.1">
 
    <description url="http://sevntu-checkstyle.github.io/sevntu.checkstyle/">
       http://sevntu-checkstyle.github.io/sevntu.checkstyle/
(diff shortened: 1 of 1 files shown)


eclipsecs-sevntu-plugin-feature/feature.xml
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<feature id="com.github.sevntu.checkstyle.checks.feature" label="Extension for eclipse-cs plugin with additional Checks" version="1.44.1">
+<feature
+      id="com.github.sevntu.checkstyle.checks.feature"
+      label="Extension for eclipse-cs plugin with additional Checks"
+      version="1.44.1">
```
rnveach, line 5: I can only assume the other changes are valid since I don't know, but I noticed when changing the formatting of these line were undone during a release by the release process. I don't know why they were, but I assume they have to remain as they were before unless we can disable the process that re-modified this file.

See https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/960
The release command that did this is `mvn -e --no-transfer-progress clean install -f eclipse-pom.xml -Plocal-deploy`.

eclipsecs-sevntu-plugin-feature/feature.xml
```diff
@@ -15,11 +18,14 @@ Eclipse Public License - Version 1.0
...
    <url>
-      <discovery label="Sources on GitHub" url="https://github.com/sevntu-checkstyle/sevntu.checkstyle"/>
-      <discovery label="SevNTU checks site" url="http://sevntu-checkstyle.github.io/sevntu.checkstyle/"/>
-      <discovery label="Eclipse Update site" url="http://sevntu-checkstyle.github.io/sevntu.checkstyle/update-site/"/>
+      <discovery label="Sevntu checkstyle update site" url="http://sevntu-checkstyle.github.io/sevntu.checkstyle/update-site/"/>
    </url>
 
-   <plugin id="eclipsecs-sevntu-plugin" download-size="0" install-size="0" version="1.44.1" unpack="false"/>
+   <plugin
+         id="eclipsecs-sevntu-plugin"
```
rnveach, line 25: Same with splitting this among multiple lines.

rnveach:
Test failure is commit message. I would rename commit to `infra: removed broken discovery sites` and put any extra text you put in the commit message to the PR's first post.

PR #1181, opened 2026-07-13
config: suppress new check JavaLineLengthCheck

DakshRJain737:
Supress new check JavaLineLengthCheck for DiamondOperatorForVariableDefinitionCheck and RequireFailForTryCatchInJunitCheck untill support for javadoc added.
Related to : https://github.com/checkstyle/checkstyle/pull/19009

### DIFF
--- a/sevntu-checks/config/suppressions.xml
+++ b/sevntu-checks/config/suppressions.xml
@@ -48,4 +48,10 @@
     <!-- Suppress MatchXpath from checkstyle#19168 -->
     <suppress checks="MatchXpath" files=".*[\\/]CommitValidationTest\.java"/>
 
+    <!-- suppressed for javadoc exceeding max length -->
+    <suppress checks="JavaLineLengthCheck"
+              files=".*[\\/]src[\\/]main[\\/]java[\\/]checkstyle[\\/]checks[\\/]coding[\\/]RequireFailForTryCatchInJunitCheck\.java"/>
+    <suppress checks="JavaLineLengthCheck"
+              files=".*[\\/]src[\\/]main[\\/]java[\\/]checkstyle[\\/]checks[\\/]coding[\\/]DiamondOperatorForVariableDefinitionCheck\.java"/>
+
 </suppressions>


sevntu-checks/config/suppressions.xml
```diff
@@ -48,4 +48,10 @@
     <!-- Suppress MatchXpath from checkstyle#19168 -->
     <suppress checks="MatchXpath" files=".*[\\/]CommitValidationTest\.java"/>
 
+    <!-- suppressed for javadoc exceeding max length -->
+    <suppress checks="JavaLineLengthCheck"
+              files=".*[\\/]src[\\/]main[\\/]java[\\/]checkstyle[\\/]checks[\\/]coding[\\/]RequireFailForTryCatchInJunitCheck\.java"/>
+    <suppress checks="JavaLineLengthCheck"
+              files=".*[\\/]src[\\/]main[\\/]java[\\/]checkstyle[\\/]checks[\\/]coding[\\/]DiamondOperatorForVariableDefinitionCheck\.java"/>
```
romani, line 55: can we simply fix violation ?
whole idea of Check to let users fix code, not to just keep suppressing.
sounds like simple update on wrapping should be done.

romani (changes requested):
thanks a lot, but

sevntu-checks/config/suppressions.xml
DakshRJain737: Sure will do it

DakshRJain737: @romani I do not find a way how to wrap them.  
```
* This Check highlights variable definition statements where <a href=
 * "http://docs.oracle.com/javase/7/docs/technotes/guides/language/type-inference-generic-instance-creation.html">
 * diamond operator</a> could be used.<br>
 * <b>Rationale</b>: using diamond operator (introduced in Java 1.7) leads to shorter code<br>
```

These are hyperlinks that are raising errors.

romani: Now you feel a pain that uses will feel after new a check run on sources.
Take a look how it is resolved in general line length Check.
We have a lot of links source code and they are not violation

PR #1184, opened 2026-07-31
config: Fixes html entities

Aman-Baliyan:
from https://github.com/checkstyle/checkstyle/pull/20397 
Fixes violations of PreferLiteralJavadocInlineTagCheck

### DIFF
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/LogicConditionNeedOptimizationCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/LogicConditionNeedOptimizationCheck.java
@@ -29,7 +29,8 @@ import com.puppycrawl.tools.checkstyle.api.TokenTypes;
  * methods and instanceof in '&amp;&amp;' and '||' conditions.
  * </p>
  * <p>
- * For example: if(getProperty() &amp;&amp; property) ==&gt; if(property &amp;&amp; getProperty()),
+ * For example:
+ * {@code if(getProperty() && property) ==> if(property && getProperty())}
  * and similarly for any expression.
  * </p>
  *
--- a/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MultipleStringLiteralsExtendedCheck.java
+++ b/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MultipleStringLiteralsExtendedCheck.java
@@ -47,7 +47,7 @@ public class MultipleStringLiteralsExtendedCheck extends AbstractCheck {
     public static final String MSG_KEY = "multiple.string.literal";
 
     /**
-     * The found strings and their positions. &lt;String, ArrayList&gt;, with
+     * The found strings and their positions. {@code <String,ArrayList>}, with
      * the ArrayList containing StringInfo objects.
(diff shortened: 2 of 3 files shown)


sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/LogicConditionNeedOptimizationCheck.java
```diff
@@ -29,7 +29,8 @@
  * methods and instanceof in '&amp;&amp;' and '||' conditions.
  * </p>
  * <p>
- * For example: if(getProperty() &amp;&amp; property) ==&gt; if(property &amp;&amp; getProperty()),
+ * For example:
+ * if(getProperty() &amp;&amp; property) =={@literal >} if(property &amp;&amp; getProperty()),
```
romani: ideally this should be code.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MultipleStringLiteralsExtendedCheck.java
```diff
@@ -47,7 +47,7 @@ public class MultipleStringLiteralsExtendedCheck extends AbstractCheck {
     public static final String MSG_KEY = "multiple.string.literal";
 
     /**
-     * The found strings and their positions. &lt;String, ArrayList&gt;, with
+     * The found strings and their positions. {@literal <}String, ArrayList{@literal >}, with
```
romani: ideally this should be code.

romani (changes requested):
items:

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MultipleStringLiteralsExtendedCheck.java
Aman-Baliyan: > ideally this should be code.

But there is not such a way to find whether to use code or literal because these docs are are not written inside `<code>` or `<pre>` .  These are written as simple text.
But we can change the violation message for it - `Prefer Javadoc inline tag '{@literal >}' or '{@code}' over '&gt;'`  and the rest depend on the users.

romani: Yes, exactly. So user will decide is it good for code or just literal.

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/LogicConditionNeedOptimizationCheck.java
Aman-Baliyan: > ideally this should be code.

done

sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/MultipleStringLiteralsExtendedCheck.java
Aman-Baliyan: > ideally this should be code.

done

romani (approved):
Thanks a lot

